PR #2, opened 2018-03-05
TAI-3717 - Update fieldsets

adamliptrot-oc:


app/views/can_we_contact_by_phone.scala.html
```diff
@@ -56,18 +56,18 @@
...
             ('_groupClass, "inline"),
             ('_fieldsetAttributes, "id=yesNoChoice" ),
-            ('_legendClass, "visuallyhidden"),
+            ('_legendSpanClass, "visually-hidden"),
             ('_radioOption1Attributes, Html("""data-journey-click="radio - select:Can we contact you by phone?:Yes" data-journey-val="Yes" data-dynamic-fields-hide="false" data-dynamic-fields="yesNoTextEntry"""") ),
             ('_radioOption2Attributes, Html("""data-journey-click="radio - select:Can we contact you by phone?:No" data-journey-val="No" data-dynamic-fields-hide="true" data-dynamic-fields="yesNoTextEntry"""") )
         )}
 
-        <fieldset id="telephoneNumberEntry-container" class="@if(yesNoForm(YesNoTextEntryForm.YesNoChoice).value.getOrElse("") != YesNoTextEntryForm.YesValue){js-hidden} data-yesNoTextEntry inline form-field-group panel-indent">
+        <div id="telephoneNumberEntry-container" class="@if(yesNoForm(YesNoTextEntryForm.YesNoChoice).value.getOrElse("") != YesNoTextEntryForm.YesValue){js-hidden} data-yesNoTextEntry inline form-field-group panel-indent">
```
markhunter27: I was considering amendment to the helper to accommodate the extra (conditional) input and place it within the fieldset since it's related content. I didn't make the change as part of the last ticket due to time constraints, but do you think it's worth doing now? - or is this less of a priority?

adamliptrot-oc: Ideally the additional fields should be right after the related field. However I think we'd be better off looking at splitting this into its components (fieldset, single radio, single input) and how we can compose those to achieve the same effect. I've raised a separate ticket for this.

app/views/incomes/payPeriod.scala.html
```diff
@@ -33,16 +33,18 @@
         preHeadingText = Messages("tai.payPeriod.preHeading"))}
 
 @howOftenPanel = @{
+    def errorMessage = Messages("tai.payPeriod.error.form.incomes.other.mandatory")
```
markhunter27: Could make these val?

app/views/can_we_contact_by_phone.scala.html
markhunter27: Sure - no prob.

PR #4, opened 2018-03-06
TAI 3666 Implemented update or remove decision page

lewey:


app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -0,0 +1,109 @@
...
+  def employmentService: EmploymentService
+  def journeyCacheService: JourneyCacheService
+  def trackingJourneyCacheService: JourneyCacheService
+
+  def decision: Action[AnyContent] = authorisedForTai(taiService).async {
+    implicit user =>
+      implicit taiRoot =>
+        implicit request =>
+          ServiceCheckLite.personDetailsCheck {
+            journeyCacheService.currentCache flatMap { currentCache =>
```
yasaracar: You can use a for comprehension instead of 2 flatmaps. It would read better

PR #8, opened 2018-03-07
TAI-3678 : Refine back links throughout service to emulate browser back.

markhunter27:


app/views/main.scala.html
```diff
@@ -152,9 +154,23 @@
         <script type="text/javascript">@gaCallsAdditionalJs.get</script>
     }
 
-    <script>
-        /* Temp - pending AF update to > 3.0.2 */
+    <script type="text/javascript">
```
adamliptrot-oc: Can this go into an external js file rather than sitting in the template? That way it will get cached and will be easily found for future maintenance.

markhunter27: Yeah I can move it into tai.js, or another new script file. The `scriptElems` content of TAAS also sits at the foot of body  .. so should be fine in terms DOM content loaded.
Will make change.

PR #11, opened 2018-03-08
TAI 3666 - Fix for the word benefit appearing multiple times on page heading.

lewey:


app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -73,11 +74,15 @@ trait CompanyBenefitController extends TaiBaseController
             } yield {
               employment match {
                 case Some(employment) =>
-                  val cache = Map(EndCompanyBenefit_EmploymentNameKey -> employment.name)
+
+                  val formattedBenefitName = (raw"(B|b)enefit(s)?".r replaceAllIn(
+                    Messages(s"tai.taxFreeAmount.table.taxComponent.${currentCache(EndCompanyBenefit_BenefitTypeKey)}"),"")).trim
```
yasaracar: I think view logic should be in view, viewModels or viewHelpers depending on the situation, but not in controllers

PR #15, opened 2018-03-12
TAI-3759 amended ApplicationConfig file taxYouPaidStatus val to start…

krishnaborra:
… using new url as per TAXCALC-FRONTEND service change

### DIFF
--- a/app/uk/gov/hmrc/tai/config/ApplicationConfig.scala
+++ b/app/uk/gov/hmrc/tai/config/ApplicationConfig.scala
@@ -18,9 +18,12 @@ package uk.gov.hmrc.tai.config
 
 import play.api.Play._
 import uk.gov.hmrc.play.config.ServicesConfig
+import uk.gov.hmrc.tai.model.tai.TaxYear
 import views.html.helper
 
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/tai/config/ApplicationConfig.scala
```diff
@@ -18,10 +18,13 @@ package uk.gov.hmrc.tai.config
 
 import play.api.Play._
 import uk.gov.hmrc.play.config.ServicesConfig
+import uk.gov.hmrc.tai.model.tai.TaxYear
 import views.html.helper
 
 object ApplicationConfig extends ServicesConfig {
 
+  val statusRange = s"${TaxYear().prev.year}-${TaxYear().prev.year + 1}"
```
grijesh-saini: please use below statement instead of val and TaxYear().prev.year + 1 
`def statusRange = s"${TaxYear().prev.year}-${TaxYear().year}" `

grijesh-saini: Is it possible to write a test case for it ?

PR #23, opened 2018-03-15
TAI-3660 - Populating CY APD screen without Session Data

grijesh-saini:


app/controllers/YourIncomeCalculationControllerNew.scala
```diff
@@ -0,0 +1,96 @@
+/*
```
RaminEsfandiari: Can you rename this to YourIncomeCalculationController? Or do you have a toggle to switch between the two? If so i am guessing you will be renaming this after you are done with the toggle?

app/uk/gov/hmrc/tai/viewModels/YourIncomeCalculationViewModelNew.scala
```diff
@@ -0,0 +1,73 @@
...
+      taxCodeIncome.status,
+      realTimeStatus,
+      latestPaymentDetails(employment),
+      employment.endDate,
+      taxCodeIncome.componentType == PensionIncome)
+  }
+
+  private def latestPaymentDetails(employment: Employment) = {
+    for {
+      latestAnnualAccount <- employment.latestAnnualAccount
```
RaminEsfandiari: Why are we calling this?

app/controllers/YourIncomeCalculationControllerNew.scala
grijesh-saini: Yes, Ramin, Currently we are giving its name as New, once we finished our code then we will delete the existing one and rename the New one to proper class name.

PR #24, opened 2018-03-15
Tai 3703 - Use minus sign encoding

gsmanderson:


test/uk/gov/hmrc/tai/util/ViewModelHelperSpec.scala
```diff
@@ -35,7 +35,7 @@ class ViewModelHelperSpec extends PlaySpec with ViewModelHelper with DateFormatC
     }
     "return the string representation of the provided MoneyPounds with a pound symbol prefix and negative sign" when {
       "the value is negative" in {
-        withPoundPrefixAndSign(MoneyPounds(-1000)) mustBe "-£1,000.00"
+        withPoundPrefixAndSign(MoneyPounds(-1000)) mustBe s"${minusSign}£1,000.00"
```
yasaracar: Can we put it in constants so that we can use it everywhere?

PR #25, opened 2018-03-15
Tai 3660 - Adding message when RTI total is not equal

grijesh-saini:


app/views/incomes/yourIncomeCalculationNew.scala.html
```diff
@@ -112,6 +112,11 @@ <h2 class="heading-medium">@Messages("tai.income.calculation.heading",s"${TaxYea
                                 }
                             </tbody>
                         </table>
+
+                        @if(model.messageWhenTotalNotEqual.isDefined) {
```
yasaracar: Can  we use map instead or is this legacy code?

PR #28, opened 2018-03-16
TAI-3695 : Bring page titles in line with GDS guidelines

markhunter27:
Added suffix ` - Pay As You Earn (PAYE) - GOV.UK` to all pages.
Suffix also added to gaCustomTitle, where present.


app/views/main.scala.html
```diff
@@ -41,6 +41,10 @@
             partialRetriever: uk.gov.hmrc.play.partials.PartialRetriever
 )
 
+@serviceSuffix = @{
+    s"- ${TaiConstants.SERVICE_NAME_PUBLIC} - GOV.UK"
```
adamliptrot-oc: Any reason why this is different to the nav title? Can they both use the same config?

markhunter27: As discussed - have updated to use Messages("tai.service.navTitle") as suggested, and removed the new constant I introduced (was based upon the labelling of the PTA tile that brings the user into our service).

PR #30, opened 2018-03-21
TAI-3664 : Implement English/Welsh language toggle

markhunter27:
This ticket covers mechanics of toggle only - not the inclusion of Welsh message file.
Language toggle implemented behind a feature toggle.

### DIFF
--- a/app/controllers/i18n/TaiLanguageController.scala
+++ b/app/controllers/i18n/TaiLanguageController.scala
@@ -65,7 +65,7 @@ trait TaiLanguageController extends LanguageController with TaiBaseController
             val redirectURL = request.headers.get(REFERER).getOrElse(fallbackURL)
 
             Future.successful(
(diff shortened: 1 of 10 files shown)


app/controllers/i18n/TaiLanguageController.scala
```diff
@@ -0,0 +1,79 @@
...
+    implicit user =>
+      implicit taiRoot =>
+        implicit request =>
+          ServiceCheckLite.personDetailsCheck {
+
+            val newLanguage =
+              if (isWelshEnabled)
+                languageMap.getOrElse(language, LanguageUtils.getCurrentLang)
+              else
+                LanguageUtils.getCurrentLang
```
adamliptrot-oc: Will this allow Welsh to be on and unable to switch to English if we toggle the lang switcher off via the flag while the user is mid-session in Welsh?

markhunter27: It would yeah, the user would need to clear cookies to revert.
Since we only have two language options I can instead default to english. It's less scalable if we were ever to introduce a third language - but that's not going to happen any time soon

PR #32, opened 2018-03-21
Tai 3739 Update error functionality No CY data

craigemmerson:


app/controllers/TaxAccountSummaryController.scala
```diff
@@ -60,19 +61,26 @@ trait TaxAccountSummaryController extends TaiBaseController
...
-              TaiSuccessResponseWithPayload(taxCodeIncomes: Seq[TaxCodeIncome]) <- taxCodeIncomesFuture
-              TaiSuccessResponseWithPayload(nonTaxCodeIncome: NonTaxCodeIncome) <- nonTaxCodeIncomeFuture
+              taxSummary <- taxSummaryFuture
+              taxCodeIncomes <- taxCodeIncomesFuture
+              nonTaxCodeIncome <- nonTaxCodeIncomeFuture
               employments <- employmentsFuture
               isAnyFormInProgress <- trackingService.isAnyIFormInProgress(nino.nino)
             } yield {
+              (taxSummary, taxCodeIncomes, nonTaxCodeIncome, employments, isAnyFormInProgress) match {
+                case (TaiTaxAccountFailureResponse(message), _, _, _, _) if message.toLowerCase.contains(TaiConstants.NpsTaxAccountDataAbsentMsg) ||
```
lewey: Could there be a scenario where by an error is caught with a different message then the ones being checked for? What would happen in this case? If that is not a possibility do you really need to check message content?

PR #34, opened 2018-03-21
Tai 3721 - Amended Textarea on Update employment journey and Update previous years income journey to be accessibly for users

krishnaborra:


app/views/main.scala.html
```diff
@@ -180,7 +180,8 @@
         "googleAnalytics" -> googleAnalytics,
         "gaCustomEvent" -> gaCustomEvent,
         "scriptElems" -> Seq(
-            Map("url" -> controllers.routes.AssetsController.versioned("javascript/tai.js"))
+            Map("url" -> controllers.routes.AssetsController.versioned("javascript/tai.js")),
+            Map("url" -> controllers.routes.AssetsController.versioned("javascript/char-count.js"))
```
markhunter27: Should the order be reversed here, as tai is dependent upon char-count script?
I think the jquery document ready (where the new code is placed within tai.js)  will fire once the DOM is loaded

PR #35, opened 2018-03-22
TAI-3699: Use the correct £ symbol

RaminEsfandiari:


app/views/incomes/bonusPaymentAmount.scala.html
```diff
@@ -51,6 +51,9 @@
     if(bonusPeriod == "year"){
         Messages("tai.bonusPaymentsAmount.year.message")
     }
+    else{
```
grijesh-saini: Do we need this else condition?

RaminEsfandiari: This is to resolve a bug where if bonus period is not equal year the input field would try to toString  the hint variable and returns empty parenthesis for some reason.

PR #41, opened 2018-03-26
TAI-3714 Remove tables

adamliptrot-oc:


app/assets/sass/modules/_navigation.scss
```diff
@@ -136,9 +135,9 @@ ul.four-buckets {
     color: #0B0C0C;
 }
 
-li a span:first-child {
-    text-decoration: underline;
-}
```
RaminEsfandiari: Can these be removed instead of being commented out maybe?

PR #46, opened 2018-03-29
Tai 3765 - Remove back link and add specific link

jordan-armstrong:


app/views/includes/ptaHeader.scala.html
```diff
@@ -22,16 +22,17 @@
   backLinkGaEventAction: Option[String] = None,
   mainHeadingText: String,
   preHeadingText: String,
-  preHeadingAccessibleAnnouncement: String = "")(implicit m: Messages)
+  preHeadingAccessibleAnnouncement: String = "",
+  backLinkContent: Option[Html] = None
```
markhunter27: Might be good to name the new parameter something like `customBackLinkContent` - to clarify that it doesn't need to be populated in the general case where a standard back link is sufficient.

conf/messages
```diff
@@ -58,7 +58,8 @@ tai.CurrentTaxYear=Current tax year
 tai.NextTaxYear=Next tax year
 tai.notKnown.response=I do not know
 tai.returnToIncomeTaxSummary=Return to your Income Tax Summary
-tai.returnToChooseTaxYear=Return to Choose Tax Year
+tai.returnToChooseTaxYear=Choose tax year
+tai.returnToChooseTaxYearHidden=Return to
```
markhunter27: Doesn't look like this is being used anywhere?

PR #51, opened 2018-04-06
TAI-3777 : Remove SessionData dependency from WDYWTD page

markhunter27:
Replace SessionData retrieval with tax account summary and employment API calls.
Revisions to resulting error trapping / failure response trapping.


test/controllers/ErrorPagesHandlerSpec.scala
```diff
@@ -358,14 +299,16 @@ class ErrorPagesHandlerSpec extends PlaySpec
   }
 
   implicit val messagesApi: MessagesApi = app.injector.instanceOf[MessagesApi]
+  implicit val tl = MockTemplateRenderer
```
yasaracar: can we name this more definitive please?

PR #61, opened 2018-04-16
TAI-3794 : Add remaining Welsh messages to service

markhunter27:
Including JS mods to enable language switching on text entry char counter.


app/assets/js/char-count.js
```diff
@@ -251,23 +251,26 @@
...
+            countNoun = ((remainingNumber === -1 || remainingNumber === 1) ? 'cymeriad' : 'o gymeriadau')
+        }
+    } else {
+        prefixClause = "You have "
+        countVerb = (remainingNumber < 0) ? 'too many' : 'remaining'
+        if (options && options.wordCount) {
+            countNoun = ((remainingNumber === -1 || remainingNumber === 1) ? 'word' : 'words')
+        } else {
+            countNoun = ((remainingNumber === -1 || remainingNumber === 1) ? 'character' : 'characters')
+        }
```
Gaur4vGaur: The logic to form the whole message is repeated, can we try to combine the messages using inline if-else? We can check which one looks betters and chose the appropriate solution.

PR #63, opened 2018-04-17
TAI-3798 - Added links to iforms at the bottom of the tax free amount…

jordan-armstrong:
… page

### DIFF
--- a/app/views/taxFreeAmountNew.scala.html
+++ b/app/views/taxFreeAmountNew.scala.html
@@ -69,7 +69,8 @@
 
                 </section>
 
-                <h2 id="addSomethingMissingHeading" class="heading">@Messages("tai.incomeTaxSummary.addMissingIncome.section.heading")</h2>
+                <section id="addSomethingMissingSection" class="section">
+                <h2 id="addSomethingMissingHeading">@Messages("tai.incomeTaxSummary.addMissingIncome.section.heading")</h2>
 
                 <p class="icon icon-plus display-block">
                     @Link.toInternalPage(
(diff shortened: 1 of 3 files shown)


conf/messages
```diff
@@ -442,7 +442,7 @@ tai.taxFreeAmount.summarysection.taxFreeExplanation=Your tax-free amount is how
 tai.taxFreeAmount.summarysection.heading=Your annual tax-free amount is
 tai.taxFreeAmount.summarysection.p2=It can be:
 tai.taxFreeAmount.summarysection.bullet1=increased if you qualify for certain allowances or tax reliefs
-tai.taxFreeAmount.summarysection.bullet2=decreased to collect tax due for things like company benefits, rental income or bank and building society interest
+tai.taxFreeAmount.summarysection.bullet2=decreased to collect tax due for company benefits or income from other sources
```
markhunter27: I'll get these in the next welsh translation.

app/views/taxFreeAmountNew.scala.html
```diff
@@ -70,6 +69,35 @@ <h2 class="flush--bottom">@Messages("tai.taxFreeAmount.detailsection.heading")</
 
                 </section>
 
+                <h2 id="addSomethingMissingHeading" class="heading">@Messages("tai.incomeTaxSummary.addMissingIncome.section.heading")</h2>
```
markhunter27: Wondering if we could place the new links within the taxFreeAmountDetail <section>, as they seem to relate to the detail items semantically.
That would mean everything would be contained within one of the two existing HTML sections - summary or detail. The new header could be switched to an h3 with a class of 'heading-medium' - so it would look the same as you have it visually, but sit semantically along with the others - allowance, additions, deductions, total, something missing, and the screen reader would treat it the same, if the user navigates into the detail section.
Not a right or wrong thing - just wondering about semantics/groupings. If you reckon it's cleaner outside of the detail, no prob.

PR #64, opened 2018-04-18
TAI-3781- Fixed multiple HOD calls issue

grijesh-saini:


app/controllers/TaxAccountSummaryController.scala
```diff
@@ -52,17 +52,12 @@ trait TaxAccountSummaryController extends TaiBaseController
...
-
             for {
-              taxSummary <- taxSummaryFuture
-              taxCodeIncomes <- taxCodeIncomesFuture
-              nonTaxCodeIncome <- nonTaxCodeIncomeFuture
-              employments <- employmentsFuture
+              taxSummary <- taxAccountService.taxAccountSummary(nino, TaxYear())
+              taxCodeIncomes <- taxAccountService.taxCodeIncomes(nino, TaxYear())
+              nonTaxCodeIncome <- taxAccountService.nonTaxCodeIncomes(nino, TaxYear())
+              employments <- employmentService.employments(nino, TaxYear())
```
Gaur4vGaur: Is it possible to prevent employments and isAnyFormInProgress  calls
Can we refactor the match statement in yield clause as employments and isAnyFormInProgress is never compared?

PR #72, opened 2018-04-24
Tai 3780 refactor remove old domain data code

Gaur4vGaur:


app/controllers/ServiceChecks.scala
```diff
@@ -16,66 +16,13 @@
...
-      case Left("deceased") => Future.successful(Redirect(routes.DeceasedController.deceased()))
-      case Left("mci") => Future.successful(Redirect(routes.ServiceController.gateKeeper()))
-      case Left("ceased") => Future.successful(Redirect(routes.WhatDoYouWantToDoController.whatDoYouWantToDoPage()))
-      case Right(details:TaxSummaryDetails) => custom.getOrElse(throw new Exception("Method not defined"))(details)
-    }
-  }
-}
+import scala.concurrent.Future
 
 object ServiceCheckLite extends TaiBaseController {
```
markhunter27: Wonder if we should rename this, now that it's the only one?

app/uk/gov/hmrc/tai/service/TaiService.scala
```diff
@@ -37,78 +28,6 @@ trait TaiService {
```
markhunter27: We could rename this one to PersonService now that it's the only method?

app/uk/gov/hmrc/tai/util/CeasedEmploymentHelper.scala
```diff
@@ -17,64 +17,15 @@
 package uk.gov.hmrc.tai.util
```
markhunter27: Possibly rename to something generic? - now that all the employment specific stuff is gone
I think the CommonVMHelper could also be removed now - and the single utility method moved into somewhere in util. Looks like it's the last trace of the 'VM' stuff ?

app/uk/gov/hmrc/tai/model/tai/JsonExtra.scala
```diff
@@ -16,11 +16,8 @@
 
 package uk.gov.hmrc.tai.model.tai
 
-import org.slf4j.Logger
 import play.api.libs.json._
 
-import scala.util._
-
 object JsonExtra {
```
markhunter27: Could move this one into the util package
The 'tai' package seems a bit empty now .. wonder if we could move the TaiRootModel into here .. or just move TaxYear into the root 'model' package and remove this sub package

markhunter27:
Our controllers are the only scala-based app components that sit outside of the 'uk.gov.hmrc.tai' package. Should we move them into the package alongside other components, now that we're tidying generally?

PR #77, opened 2018-04-27
Tai 3780 refctr - code refactor and increase the code coverage

Gaur4vGaur:


app/views/benefits/removeCompanyBenefitStopDate.scala.html
```diff
@@ -18,7 +18,7 @@
 @import uk.gov.hmrc.urls.Link
 @import includes.{ptaHeader, cancelLink}
 @import uk.gov.hmrc.tai.forms.benefits.RemoveCompanyBenefitStopDateForm
-@import includes.forms.{inputRadioGroupNewStyling, submit}
+@import includes.forms.{inputRadioGroupStyling, submit}
```
markhunter27: I think we could drop the 'styling' suffix entirely. It was originally referring to the newer radio button styling, as per 
http://govuk-elements.herokuapp.com/form-elements/#form-radio-buttons
.. but since we've now removed the old stuff I don't think we need it.

Gaur4vGaur: I will address it in my next request

PR #85, opened 2018-05-09
Tai 3800 increased test coverage to 93.22%

craigemmerson:


test/controllers/income/bbsi/BbsiCloseAccountControllerSpec.scala
```diff
@@ -406,6 +408,42 @@ class BbsiCloseAccountControllerSpec extends PlaySpec
 
       status(result) mustBe BAD_REQUEST
     }
+    "no name provided" in {
```
markhunter27: The two new tests are verifying an internal server error response, so I think they should be moved into a test block of their own - at the moment they're sitting inside a block which states "return bad request"

test/uk/gov/hmrc/tai/service/benefits/CompanyCarServiceSpec.scala
```diff
@@ -149,6 +149,14 @@ class CompanyCarServiceSpec extends PlaySpec
         val expectedResult = TaiNoCompanyCarFoundResponse("No company car found")
         Await.result(sut.beginJourney(generateNino, 1), 5 seconds) mustBe expectedResult
       }
+      "there are no employments" in  {
```
markhunter27: This new test is verifying that a runtime exception is thrown, so I think it needs to sit outside of the current test block which is grouping all situations that return a TaiNoCompanyCarFoundResponse

PR #86, opened 2018-05-09
TAI-3811 : Integrate with new TAI API for Person details retrieval

markhunter27:
New operation now exposed on PersonService - but not yet used anywhere.
TAI-3814 on new sprint will replace all call to the old TaiRoot method, with this one.


app/uk/gov/hmrc/tai/model/domain/Person.scala
```diff
@@ -0,0 +1,44 @@
...
+                  dateOfBirth: Option[LocalDate],
+                  address: Address,
+                  isDeceased: Boolean,
+                  hasCorruptData: Boolean)
+
+object Person {
+  implicit val personFormat: Format[Person] = Json.format[Person]
+}
+
+case class Address(line1: String,
```
yasaracar: Do we use address in the FE? If not we shouldn't have this in the model

yasaracar: We can remove this during or after the second pr

markhunter27: Yeah - will remove Address, and DOB fields on the latest PR.

PR #87, opened 2018-05-10
TAI-3752 - Persist user data in form controls for 'Remove Employment'

zachhall1234:


boseabhishek:
PR reviewed .. looks good!

app/uk/gov/hmrc/tai/service/JourneyCacheService.scala
```diff
@@ -88,6 +88,21 @@ trait JourneyCacheService extends JourneyCacheConstants {
     }
   }
 
+  def collectedOptionalValues(optionalValues: Seq[String])(implicit hc: HeaderCarrier): Future[Seq[Option[String]]] = {
```
yasaracar: Can you write unit tests for this method please?

app/controllers/employments/EndEmploymentController.scala
```diff
@@ -195,11 +195,18 @@ trait EndEmploymentController extends TaiBaseController
         implicit request =>
           val nino = Nino(user.getNino)
           ServiceCheckLite.personDetailsCheck {
-            employmentService.employment(nino, employmentId).map {
-              case Some(employment) =>
+            for{
+              employment <- employmentService.employment(nino, employmentId)
+              cache <- journeyCacheService.currentValueAsDate(EndEmployment_EndDateKey)
+            }yield (employment, cache) match{
```
yasaracar: Do we have unit tests or each case?

app/uk/gov/hmrc/tai/service/JourneyCacheService.scala
markhunter27: For consistency with the `mandatoryValues()` equivalent, I reckon we should also accept varargs param, and name it just 'optionalValues` - otherwise we've got an inconsistent interface on the JCS.
You could also call your new method from within collectedValues now, and remove second half of that method - since it's basically your method - would tidy your code (not essential though :)

PR #94, opened 2018-05-15
Tai 3760 - Persist user data in form controls - ADD A MISSING PENSION DETAILS

richard-poole:


app/controllers/pensions/AddPensionProviderController.scala
```diff
@@ -70,7 +70,14 @@ trait AddPensionProviderController extends TaiBaseController
     implicit person =>
       implicit request =>
         ServiceCheckLite.personDetailsCheck {
-          Future.successful(Ok(views.html.pensions.addPensionName(PensionProviderNameForm.form)))
+          journeyCacheService.currentValue(AddPensionProvider_NameKey) map {
+            case Some(pensionName) => {
+              Ok(views.html.pensions.addPensionName(PensionProviderNameForm.form.fill(pensionName)))
+            }
```
craigemmerson: to make it cleaner we can get rid of the match here, as its opional we can just return the page and fill with the name, if its not there its empty string.

PR #95, opened 2018-05-15
TAI-3815 refactor of doYouGetThisPension method

craigemmerson:


app/controllers/pensions/UpdatePensionProviderController.scala
```diff
@@ -81,20 +82,28 @@ trait UpdatePensionProviderController extends TaiBaseController
...
         }
   }
 
+  private def cacheAndCreateView(id: Int, taxCodeIncome: TaxCodeIncome)(implicit hc: HeaderCarrier,
+                                                                        request: Request[AnyContent],
+                                                                        user: TaiUser): Future[Result] = {
+    for {
+      _ <- journeyCacheService.cache(Map(UpdatePensionProvider_IdKey -> id.toString,
+        UpdatePensionProvider_NameKey -> taxCodeIncome.name))
+      optionalYesNo <- journeyCacheService.currentValue(UpdatePensionProvider_ReceivePensionQuestionKey)
```
markhunter27: The `journeyCacheService.cache()` method will return the updated cache in full, so you could skip the extra hit on journey cache here - no need to go back and ask for currentValue.

PR #107, opened 2018-06-18
St 9

Froom2:


app/uk/gov/hmrc/tai/viewModels/PreviousYearUnderpaymentViewModel.scala
```diff
@@ -0,0 +1,55 @@
...
+object PreviousYearUnderpaymentViewModel extends ViewModelHelper {
+
+  def apply(codingComponents: Seq[CodingComponent], employments: Seq[Employment]): PreviousYearUnderpaymentViewModel = {
+
+    val taxYear = TaxYear().prev
+
+    val actuallyPaid = (for {
+      emp <- employments
+      account <- emp.annualAccounts.find(_.taxYear.year == taxYear.year)
+    } yield account.totalTaxPaidYearToDate).foldLeft(BigDecimal(0.00))(_ + _)
```
amaalali: Could this just be a `reduce` since we aren't really using the seed value for fold?

test/controllers/PreviousYearUnderpaymentControllerSpec.scala
```diff
@@ -0,0 +1,90 @@
...
+import uk.gov.hmrc.play.frontend.auth.connectors.{AuthConnector, DelegationConnector}
+import uk.gov.hmrc.play.partials.FormPartialRetriever
+import uk.gov.hmrc.renderer.TemplateRenderer
+import uk.gov.hmrc.tai.service.{AuditService, CodingComponentService, EmploymentService, PersonService}
+import uk.gov.hmrc.tai.viewModels.PreviousYearUnderpaymentViewModel
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+class UnderPaymentFromPreviousYearControllerSpec extends PlaySpec
```
amaalali: Should we add a test for what the controller returns? Or that the template is passed the `PreviousYearUnderpaymentViewModel` that we expect. Thoughts? Would this add value?

app/controllers/PotentialUnderpaymentController.scala
```diff
@@ -49,7 +49,7 @@ trait PotentialUnderpaymentController extends TaiBaseController
     implicit user =>
       implicit person =>
         implicit request =>
-          ServiceCheckLite.personDetailsCheck {
+          val test = ServiceCheckLite.personDetailsCheck {
```
amaalali: I think that this variable might have been left in while debugging?

test/views/html/potentialUnderpaymentViewSpec.scala
```diff
@@ -31,6 +31,10 @@ class potentialUnderpaymentViewSpec extends TaiViewSpec {
 
   "the Potential Underpayment view method" should {
 
+    "test" in {
```
amaalali: Is this left over from debugging?

PR #140, opened 2018-07-30
ST-117: What happens next

lewey:


test/controllers/TaxCodeChangeControllerSpec.scala
```diff
@@ -18,10 +18,11 @@ package controllers
 
 import builders.{AuthBuilder, RequestBuilder}
 import mocks.MockTemplateRenderer
+import org.jsoup.Jsoup
 import org.mockito.Matchers.{any, eq => mockEq}
 import org.mockito.Mockito.when
 import org.scalatest.mock.MockitoSugar
-import org.scalatestplus.play.PlaySpec
+import org.scalatestplus.play.{OneAppPerSuite, PlaySpec}
```
amaalali: We have some imports that don't seem to be used anymore: eg `OneAppPerSuite` and `import org.mockito.Matchers.{eq => mockEq}
`

conf/messages
```diff
@@ -1373,6 +1373,20 @@ tai.taxCollected.atSource.otherIncome.description=Tax on other income not collec
 tai.taxCollected.byPAYE.title=Tax collected through your tax code
 tai.taxCollected.byPAYE.description=Pay As You Earn (PAYE) does not include tax paid elsewhere. The tax you’ll pay through PAYE is £{0}.
 
+
+# tax code change
+taxCode.change.journey.preHeading= Your tax code change
+taxCode.change.whatHappensNext.title=What happens next
```
RaminEsfandiari: There are already three instances of this text in the messages file. So I would say it's best to move all 4 (including yours) into under one key and refactor across. Same goes for Welsh. This is only a suggestion since I think having page agnostic keys helps reduce duplication and increase readability when it comes to using it in views, and spec files.

conf/messages
```diff
@@ -1373,6 +1373,20 @@ tai.taxCollected.atSource.otherIncome.description=Tax on other income not collec
...
+taxCode.change.whatHappensNext.paragragh1=If the information about your tax code is right, you do not need to do anything.
+taxCode.change.whatHappensNext.paragragh2=You can check {0} to see how much tax we think you will pay.
+taxCode.change.whatHappensNext.paragragh3=You can {0} in your PAYE Income Tax summary. If this affects your tax code, you will get a new one.
+taxCode.change.whatHappensNext.yourIncomeTaxEstimate.link=your Income Tax estimate
+taxCode.change.whatHappensNext.updateCurrentIncomeOrBenefits.link=update your current income or benefits
+taxCode.change.whatHappensNext.wrongInformation.text=If the information you have seen is wrong
+taxCode.change.yourTaxFreeAmount.title=Your tax-free amount
```
RaminEsfandiari: Already in messages file

test/views/html/taxCodeChange/WhatHappensNextViewSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  "what happens next" should {
+    behave like pageWithTitle(messages("taxCode.change.whatHappensNext.title"))
+    behave like pageWithBackLink
+    behave like pageWithCombinedHeader(messages("taxCode.change.journey.preHeading"), messages("taxCode.change.whatHappensNext.title"))
+
+    "display static messages" in {
+      doc must haveParagraphWithText(messages("taxCode.change.whatHappensNext.paragragh1"))
+
+      doc(view).select("#check-income-tax-estimate").html() mustBe Html(messages("taxCode.change.whatHappensNext.paragragh2",
```
RaminEsfandiari: Interesting choice to use 'select' over 'getElementById'. Both achieve the same result, the only difference is you won't need to prepend the id with #.

conf/messages
lewey: Discussed and will leave duplicates for the time being

conf/messages
lewey: Discussed and will leave duplicates for the time being.

PR #147, opened 2018-08-06
ST-116: Your tax free amount page

lewey:


app/controllers/TaxCodeChangeController.scala
```diff
@@ -61,10 +71,32 @@ trait TaxCodeChangeController extends TaiBaseController
...
+            for {
+              employmentNames <- employmentNameFuture
+              taxCodeChange <- taxCodeChangeFuture
+              codingComponents <- codingComponentService.taxFreeAmountComponents(nino, TaxYear())
+              companyCarBenefits <- companyCarService.companyCarOnCodingComponents(nino, codingComponents)
+
+            } yield {
+              (taxCodeChange) match {
+                case (TaiSuccessResponseWithPayload(taxCodeHistory: TaxCodeHistory)) => {
+                  taxCodeHistory.taxCodeRecord match {
```
amaalali: We have a case here for this option to be `None` which we don't handle, so we could run into a match error. Instead we can handle that and simplify the nested matching by matching in the first case on 

`TaiSuccessResponseWithPayload(TaxCodeHistory(_, Some(taxCodeRecords: Seq[TaxCodeRecord])))`

instead of 

```
case (TaiSuccessResponseWithPayload(taxCodeHistory: TaxCodeHistory)) => {
                  taxCodeHistory.taxCodeRecord match {
	                    case Some(taxCodeRecords) =>
```

add-harris:
we've made the requested changes & changed the or operator to '||'

app/uk/gov/hmrc/tai/viewModels/taxCodeChange/YourTaxFreeAmountViewModel.scala
```diff
@@ -0,0 +1,52 @@
...
+  * Created by digital032748 on 25/07/18.
+  */
+case class YourTaxFreeAmountViewModel(taxCodeDateRange: String, annualTaxFreeAmount:String, taxFreeAmountSummary: TaxFreeAmountSummaryViewModel){}
+
+object YourTaxFreeAmountViewModel extends ViewModelHelper with TaxAccountCalculator {
+
+  def apply(taxCodeRecords: Seq[TaxCodeRecord], codingComponents: Seq[CodingComponent], employmentName: Map[Int, String],
+            companyCarBenefits: Seq[CompanyCarBenefit])(implicit messages: Messages): YourTaxFreeAmountViewModel = {
+
+    val p2IssuedDate = taxCodeRecords.map(_.p2Date).sorted.last
```
amaalali: Should this logic be elsewhere? We can ask the service to return just this `taxCodeRecord`, so that service owns all logic around what shape it comes back as, as well as the knowledge that it needs to be sorted etc

app/uk/gov/hmrc/tai/viewModels/taxCodeChange/YourTaxFreeAmountViewModel.scala
```diff
@@ -0,0 +1,52 @@
...
+
+    val taxCodeDateRange = dynamicDateRangeHtmlNonBreak(LocalDate.parse(p2IssuedDate,DateTimeFormat.forPattern("yyyy-MM-dd")),
+    TaxYearResolver.endOfCurrentTaxYear)
+    val annualTaxFreeAmount = withPoundPrefixAndSign(MoneyPounds(taxFreeAmount(codingComponents), 0))
```
amaalali: We're seeing this in quite a few places where we have 

```
withPoundPrefixAndSign(MoneyPounds(<BigDecimal>, 0))
```

Is it worth just creating an overloaded method that takes a BigDecimal and does this for us? Just an idea to reduce the clutter

PR #175, opened 2018-10-02
ST-296: Remove blue heading from Estimated Pay

SeanArmstrong:
- [x] Unit tests passing
- [x] Coverage reached
- [x] Reviewed

### DIFF
--- a/app/views/incomes/confirm_save_Income.scala.html
+++ b/app/views/incomes/confirm_save_Income.scala.html
@@ -31,14 +31,14 @@ request: Request[_], messages: Messages, user: controllers.auth.TaiUser, templat
 @header = @{
     ptaHeader(
         displayBackLink = true,
-        mainHeadingText = Messages("tai.incomes.confirm.save.title"),
-        preHeadingText = Messages("tai.incomes.edit.preHeading", income.name),
+        mainHeadingText = messages("tai.incomes.confirm.save.title"),
+        preHeadingText = messages("tai.incomes.edit.preHeading", income.name),
         headerWidthClasses = Some("column-full")
     )
 }
 
 @main(
-    title = Messages("tai.incomes.confirm.save.title"),
+    title = messages("tai.incomes.confirm.save.title"),
     articleClasses = Some("full-width"),
     pageHeader = Some(header),
     user = Some(user),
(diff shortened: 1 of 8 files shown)


app/views/incomes/estimatedPay.scala.html
```diff
@@ -48,10 +48,9 @@
     employerName = Some(employerName),
     gaCustomTitle = None
 ){
-    <div class="section--branded hmrc flush--bottom soft--top">
-
-        <h2>@Messages("tai.estimatedPay.yourPay")</h2>
-        <p class="h1-heading" id="payBeforeTax">£@MoneyPounds(payBeforeTax.getOrElse(0),0).quantity</p>
+    <div>
+        <h2 class="flush--top">@Messages("tai.estimatedPay.yourPay")</h2>
```
amaalali: We should probably move away from using `Messages` directly and use `messages` which is implicitly passed to the template

app/views/incomes/estimatedPay.scala.html
```diff
@@ -64,25 +63,27 @@ <h2>@Messages("tai.estimatedPay.yourPay")</h2>
...
+                @Messages("tai.estimatedPay.whySame.text")
             }
         }
 
     </div>
 
+    <div class="section">
+
+        @includes.link(
+            copy=Messages("tai.estimatedPay.checkTaxEstimate"),
```
amaalali: Same comment on use of `Messages` as above.

app/views/incomes/editIncome.scala.html
```diff
@@ -57,8 +57,8 @@ <h2 class="h2-heading">@Messages("tai.incomes.edit.howTo.subtitle")</h2>
         </ul>
     </div>
 
-    <div class="section--branded hmrc">
-        <h2 class="h2-heading">@Messages("tai.incomes.edit.incomeTax.update.heading")</h2>
+    <div>
+        <h2>@Messages("tai.incomes.edit.incomeTax.update.heading")</h2>
```
amaalali: Use of `messages`  over `Messages`?

app/views/incomes/confirm_save_Income.scala.html
```diff
@@ -48,7 +48,7 @@
     <div class="grid-row">
         <div class="grid grid-2-3">
             <div class="inner-block">
-                <h2 class="heading-medium heading--underlined flush--top">@Messages("tai.taxYear",s"${Dates.formatDate(TaxYearResolver.startOfCurrentTaxYear)}",s"${Dates.formatDate(TaxYearResolver.endOfCurrentTaxYear)}")</h2>
+                <h2 class="flush--top">@Messages("tai.taxYear",s"${Dates.formatDate(TaxYearResolver.startOfCurrentTaxYear)}",s"${Dates.formatDate(TaxYearResolver.endOfCurrentTaxYear)}")</h2>
```
amaalali: More a discussion on this page which existed previously, but what do you think about the hardcoded things like these dates in the views? Should these dates be passed in with the view model?

amaalali:
Would it be possible to give the file `tai03092018.scss` a better name? Its a bit of a meaningless name at the moment

app/views/incomes/confirm_save_Income.scala.html
SeanArmstrong: Its probably fine when its static stuff like this but I don't know if the view should know about TaxYearResolver

Froom2:
This has been QA'd and looks fine.

app/views/incomes/confirm_save_Income.scala.html
amaalali: I was thinking along the same lines. The date is something that could easily be passed in

PR #190, opened 2018-10-15
St 308

BriWak:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed


conf/application.conf
```diff
@@ -15,7 +15,7 @@
 include "common.conf"
 
 application.global = uk.gov.hmrc.tai.config.ApplicationGlobal
-application.router=app.Routes
+play.http.router=app.Routes
```
SeanArmstrong: What this change about?

conf/tai.routes
```diff
@@ -147,8 +147,8 @@ GET         /income-tax-comparison
 GET         /your-income-calculation-details/:id                                       controllers.YourIncomeCalculationController.yourIncomeCalculationPage(id: Int)
 GET         /print-your-income-calculation-details/:id                                 controllers.YourIncomeCalculationController.printYourIncomeCalculationPage(id: Int)
 
-GET         /your-income-calculation-previous-year/:year/:empId                       controllers.YourIncomeCalculationController.yourIncomeCalculationHistoricYears(year: uk.gov.hmrc.tai.model.TaxYear, empId: Int)
-GET         /print-your-income-calculation-previous-year/:year/:empId                 controllers.YourIncomeCalculationController.printYourIncomeCalculationHistoricYears(year: uk.gov.hmrc.tai.model.TaxYear, empId: Int)
+GET         /your-income-calculation-previous-year/:year/:empId                        controllers.YourIncomeCalculationController.yourIncomeCalculationHistoricYears(year: uk.gov.hmrc.tai.model.TaxYear, empId: Int)
```
SeanArmstrong: Have you changed this file?

BriWak: It's just spacing to line everything up the same

conf/application.conf
BriWak: application.router is deprecated. We were getting warning messages when running tests advising to use play.http.router instead. It seems to do exactly the same thing, so I spoke to @RaminEsfandiari and he said it should be ok to change.

SeanArmstrong: There is an argument that things like this and the reformatting of the `conf/tai.routes` should be done in separate PRs so that it doesn't add noise to the work you are actually doing. Maybe thats being abit pedantic though when there isn't many changes

app/views/incomes/editPension.scala.html
```diff
@@ -19,70 +19,69 @@
...
+            }
+            <div class="column-half">
+                <label class="form-label form-label-bold">@messages("tai.incomes.amount.current.pension")</label>
+                <input type="hidden" id="oldAmount" name='@income("oldAmount").name' value='@income("oldAmount").value' />£@income("oldAmount").value
+            </div>
+            <div class="column-half">
+                <label class="form-label form-label-bold" id="amountLabel">@messages("tai.incomes.amount.new.pension")</label>
+                <div class="input-wrapper--currency">
+                    <span class="input-currency-symbol" aria-hidden="true">£</span>
+                    <input aria-labelledby="amountLabel" class="form-control-currency" type="text" id="newAmount" name='@income("newAmount").name' value='' />
```
SeanArmstrong: This could be in the form of something like 

`@input(income("newAmount"), isCurrency = true, fieldErrorsEnabled = false)`

So that we dont have to duplicate the classes over the two places and keeps all inputs consistent. You may have to add the aria-labelledby as a optional param

app/views/incomes/editPension.scala.html
```diff
@@ -19,70 +19,69 @@
...
+    <div class="form-field-group @if(income.hasErrors) {error} margin-bottom-large">
 
-            <div class="form-field-group">
-                <p><input type="submit" class="button" value='@messages("tai.next")'/></p>
+        <div class="grid-row">
+            @income.errors(income("newAmount").name).map{ error =>
+            <span class="error-notification">@messages(error.message)</span>
+            }
+            <div class="column-half">
+                <label class="form-label form-label-bold">@messages("tai.incomes.amount.current.pension")</label>
```
SeanArmstrong: Technically this shouldnt be a label as there is no associated text element

app/views/incomes/editPension.scala.html
```diff
@@ -19,70 +19,69 @@
...
+                <input type="hidden" id="oldAmount" name='@income("oldAmount").name' value='@income("oldAmount").value' />£@income("oldAmount").value
```
SeanArmstrong: How does this handle responsibly? There should be a gap between the old and new when stacked

PR #192, opened 2018-10-16
ST-315: EditIncome reformat entering yearly amount

SeanArmstrong:
- [x] Unit tests passing
- [x] Coverage reached
- [x] Reviewed
- [x] Welsh translation


app/views/incomes/editIncome.scala.html
```diff
@@ -66,31 +67,33 @@
...
+                @income.errors(income("newAmount").name).map{ error =>
+                    <span class="error-notification validation-error__inline-error">@messages(error.message)</span>
+                }
+
+                <div class="grid-row">
+                    <div class="column-half">
+                        <div class="form-label form-label-bold">@messages("tai.incomes.amount.current")</div>
+                        <div>
+                            <input type="hidden" id="oldAmount" name='@income("oldAmount").name' value='@income("oldAmount").value' />
+                            <p><span>£</span> @income("oldAmount").value</p>
```
BriWak: Minor but is span needed here? should work fine with `£@income("oldAmount").value`

If there's a good reason for having it, I should probably add it to the pension page as well. Everything else looks fine.

SeanArmstrong: Erm, I think I added it initially thinking that I might need to add some styling around it. But you are right there is not reason to have it unless it changes how the screen reader works

app/views/incomes/editIncome.scala.html
```diff
@@ -66,31 +67,33 @@
...
-                                <span class="error-notification validation-error__inline-error">@messages(error.message)</span>
-                            }
-                            <td class="text--right"><input type="hidden" id="oldAmount" name='@income("oldAmount").name' value='@income("oldAmount").value' />@income("oldAmount").value</td>
-                            <td class="text--right"><input aria-labelledby="amountLabel" class="text--right input--small income-data-input" type="text" id="newAmount" name='@income("newAmount").name' value='' /></td>
-                        </tr>
-                    </tbody>
-                </table>
-                <p class="form-hint-list">@Messages("tai.enterWholeAndNumericalOnly")</p>
+
+        <p>@Messages("tai.enterWholeAndNumericalOnly")</p>
```
amaalali: Can we use the implicit `messages` over `Messages`

app/views/incomes/editIncome.scala.html
```diff
@@ -66,31 +67,33 @@
...
+        <div class="form-field-group @if(income.hasErrors) {error}">
+
+                @income.errors(income("newAmount").name).map{ error =>
+                    <span class="error-notification validation-error__inline-error">@messages(error.message)</span>
+                }
+
+                <div class="grid-row">
+                    <div class="column-half">
+                        <div class="form-label form-label-bold">@messages("tai.incomes.amount.current")</div>
+                        <div>
```
amaalali: Is this necessary?

SeanArmstrong: Probably not. I will check it out good spot 👍

SeanArmstrong:
Passed QA with @vajarova

PR #201, opened 2018-10-23
ST-334 amended messages from 'fortnightly' to 'every 2 weeks'

kenneth-rayner:
- [x] Unit tests passing
- [x] Coverage reached
- [x] Reviewed


SeanArmstrong:
There may be other parts of the journey that reference the "Fortnightly"
May be work clicking it as an option and continuing through the journey. The error messages may also refer to it but these are referenced in another ticket

PR #212, opened 2018-10-25
DDCNLS-4119 changed urls and test

OktawianC:
changed url from paye/company-car/service-landing-page to paye/company-car/details


OktawianC:
it doesn't need merge, it already has been added

PR #219, opened 2018-10-30
Convert estimated pay update forms to consistent partial

SeanArmstrong:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed


SeanArmstrong:
@amaalali I was really struggling with the name

SeanArmstrong:
Its too difficult to turn into a full form at the moment due to editIncomes hidden fields. Its probably more composable outside of the form anyway

app/views/includes/compareValueInput.scala.html
```diff
@@ -0,0 +1,41 @@
...
+)(implicit m: Messages)
+
+@import includes.forms.input
+@import uk.gov.hmrc.play.views.helpers.MoneyPounds
+
+<div class="grid-row">
+    <div class="column-half">
+        <div class="form-label form-label-bold">@previousLabel</div>
+        <input type="hidden" id="oldAmount" name='@previousFieldName' value='@previousFieldValue' />
+        <p>£@MoneyPounds(previousFieldValue.toInt ,0).quantity</p>
```
amaalali: We should probably use this util here https://github.com/hmrc/tai-frontend/blob/master/app/uk/gov/hmrc/tai/util/MonetaryUtil.scala#L29

app/views/includes/compareValueInput.scala.html
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@(
+    previousLabel: String,
+    newLabel: String,
+    newField: Field,
+    previousFieldName: String,
+    previousFieldValue: String
```
amaalali: Should this be a Moneyponds instead? This could be a runtime error if the previous value is malformed with a comma, and therefore fail for `toInt`

app/views/includes/compareValueInput.scala.html
SeanArmstrong: Addressed

app/views/includes/compareValueInput.scala.html
SeanArmstrong: I think this is a wider issue where we don't sanitize/validate our values from the form correctly before saving to the cache.

PR #236, opened 2018-11-13
ST-363: CY plus one comparison page

vajarova:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


SeanArmstrong:
Think we need to have another look at how we are breaking names. That table is just not responsive enough for our needs

screenshot 2018-11-23 at 09 18 04

PR #241, opened 2018-11-14
St 369

amaalali:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed


app/controllers/auth/WithAuthorisedForTai.scala
```diff
@@ -132,4 +133,6 @@ case class TaiUser(authContext: AuthContext, person: Person){
 
     def getUTR = authContext.principal.accounts.sa.map(sa => sa.utr.utr).getOrElse("")
 
+    def nino: Nino = Nino(getNino)
```
SeanArmstrong: Are you planning on changing all of the calls of user.getNino to user.nino

app/controllers/income/UpdateIncomeNextYearController.scala
```diff
@@ -40,7 +46,7 @@ trait UpdateIncomeNextYearController extends TaiBaseController
   with FeatureTogglesConfig
   with Auditable {
 
-  val updateNextYearsIncomeService: UpdateNextYearsIncomeService
+  def updateNextYearsIncomeService: UpdateNextYearsIncomeService
```
SeanArmstrong: Why has this changed? Whats the rational for it? The override in the object is still a val

app/uk/gov/hmrc/tai/service/UpdateNextYearsIncomeService.scala
```diff
@@ -85,4 +88,16 @@ class UpdateNextYearsIncomeService {
...
     }
   }
+
+  def submit(employmentId: Int, nino: Nino)(implicit hc: HeaderCarrier): Future[TaiResponse] = {
+    get(employmentId, nino) flatMap {
+      case UpdateNextYearsIncomeCacheModel(employmentName, _, _, Some(newValue)) => {
+        taxAccountService.updateEstimatedIncome(nino, newValue, TaxYear().next, employmentId)
+      }
+      case _ => {
+        // TODO: Should this be something more meaningful?
```
SeanArmstrong: This this the sort of place where you want to redirect back to the edit page? There is not real reason why they should ever be here without a valid cache

app/uk/gov/hmrc/tai/util/TaxYearRangeUtil.scala
```diff
@@ -47,6 +49,22 @@ object TaxYearRangeUtil {
     dateRangeSingleLine(messageRangeKeyBetween, TaxYearResolver.startOfCurrentTaxYear, TaxYearResolver.endOfCurrentTaxYear)
   }
 
+  def futureTaxYearRangeHtmlNonBreak(add: Int)(implicit messages: Messages): String = {
+    @tailrec
```
SeanArmstrong: I think this can refactored and have the logic inside the TaxYear class

app/uk/gov/hmrc/tai/viewModels/income/ConfirmAmountEnteredViewModel.scala
```diff
@@ -0,0 +1,64 @@
...
+      employerName = employerName,
+      yearRange = currentYear,
+      mainText = mainParagraphText,
+      onConfirm = confirmUrl,
+      onCancel = onCancelUrl
+    )
+  }
+
+  def nextYearEstimatedPay(employmentId: Int, employerName: String, estimatedIncome: Int)(implicit messages: Messages): ConfirmAmountEnteredViewModel = {
+    val nextYearRange: String = TaxYearRangeUtil.futureTaxYearRangeHtmlNonBreak(1)
```
SeanArmstrong: I don't think this api is very readable from this side either.

test/controllers/income/UpdateIncomeNextYearControllerSpec.scala
```diff
@@ -186,6 +187,108 @@ class UpdateIncomeNextYearControllerSpec extends PlaySpec
...
+            Future.successful(serviceResponse)
+          )
+
+          val result = controller.confirm(employmentID)(fakeRequest)
+
+          status(result) mustBe INTERNAL_SERVER_ERROR
+        }
+
+
+        "respond with and Bad Request and redirect to the edit page" ignore {
```
SeanArmstrong: ignore

test/uk/gov/hmrc/tai/service/UpdateNextYearsIncomeServiceSpec.scala
```diff
@@ -161,6 +163,34 @@ class UpdateNextYearsIncomeServiceSpec extends PlaySpec with MockitoSugar {
...
+      val service = new UpdateNextYearsIncomeServiceTest
+
+      when(
+        service.journeyCacheService.currentCache(any())
+      ).thenReturn(
+        Future.successful(fullMap(employmentName, employmentId, employmentAmount))
+      )
+
+      when(
+        service.taxAccountService.updateEstimatedIncome(any(), any(), any(), any())(any())
```
SeanArmstrong: I think we should try to reduce the amount of anys we use and be more explicit on params.

app/controllers/auth/WithAuthorisedForTai.scala
amaalali: No I wasn't at first, do you think we should deprecate the `getNino` api?

app/controllers/income/UpdateIncomeNextYearController.scala
amaalali: I'm not sure there was one, I can remove this since its unnecessary

app/uk/gov/hmrc/tai/service/UpdateNextYearsIncomeService.scala
amaalali: This is the same thing I was thinking with https://github.com/hmrc/tai-frontend/pull/241/files#diff-cc93f81a2f6a026fd7936a9da58b5828R235

test/uk/gov/hmrc/tai/service/UpdateNextYearsIncomeServiceSpec.scala
amaalali: Agreed, in this case we're saying we don't care about the response from taxAccountService, only that it calls through to taxAccountService with the details

PR #248, opened 2018-11-19
ST-382 Updated error message for monetary amount fields

kenneth-rayner:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed


kenneth-rayner:
The tests do not belong in the controller as they are all testing the new amount is valid

PR #249, opened 2018-11-19
ST-364: Updated Estimated pay CY Plus 1 journey to include pensions

BriWak:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] ~Acceptance tests passing~
- [x] Reviewed


test/views/html/incomes/nextYear/updateIncomeCYPlus1StartSpec.scala
```diff
@@ -39,7 +40,16 @@ class updateIncomeCYPlus1StartSpec extends TaiViewSpec {
       doc(view).getElementsByClass("button").text must include(messages("tai.income.details.updateTaxableIncome.update"))
     }
 
+    "contain the correct content when income is from pension" in {
```
SeanArmstrong: These is a possible improvement to the tests (may be over the top).

So the main question is what happens if you accidentally removed the if statement from the view? 

I believe that all of these tests would still pass. Perhaps we should be testing the negative too:
`doc(view).getElementsByTag("p").text mustNot include(messages("tai.updateIncome.CYPlus1.start.paragraph2", employerName))`

BriWak: Done

PR #369, opened 2019-01-18
ST-380: Added logic to be applied on titles that have long words in t…

vajarova:
…he name

- [x] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/assets/sass/tai05102018.scss
```diff
@@ -77,7 +77,13 @@ hr {
 }
 
 .carry-over{
-  -ms-word-break: break-all;
+  @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
```
SeanArmstrong: What is the need for high contrast here?

app/assets/sass/tai05102018.scss
```diff
@@ -77,7 +77,13 @@ hr {
...
+    display: inline-block;
+  }
+
+  @supports (-ms-accelerator:true) {
```
SeanArmstrong: I dont think you should be using this ? What was the use case?

https://developer.mozilla.org/en-US/docs/Web/CSS/-ms-accelerator

app/assets/sass/tai05102018.scss
```diff
@@ -86,4 +92,18 @@ hr {
   -webkit-hyphens: none;
   -moz-hyphens: none;
   hyphens: none;
+}
+
+h1>span.small{
+  @media(max-width: 668px){
```
SeanArmstrong: Is there not any pre-done breakpoints here already defined?

PR #398, opened 2019-01-25
Added confidence level to new auth

oc-ste:
- [x] Unit tests passing
- [x] Coverage reached
- [x] Reviewed


app/controllers/auth/AuthAction.scala
```diff
@@ -32,7 +32,7 @@ import scala.concurrent.{ExecutionContext, Future}
 
 case class AuthenticatedRequest[A](request: Request[A], taiUser: AuthedUser) extends WrappedRequest[A](request)
 
-case class AuthedUser(name: String, validNino: String, utr: String, userDetailsUri: String) {
+case class AuthedUser(name: String, validNino: String, utr: String, userDetailsUri: String, confidenceLevel: String) {
```
lewey: Would this not be better off being a Sealed Case object

PR #467, opened 2019-03-05
ST-559 same estimated pay

oc-ste:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed

Acceptance tests: https://github.com/hmrc/tai-acceptance-test-suite/pull/129


app/controllers/IncomeController.scala
```diff
@@ -110,34 +125,49 @@ class IncomeController @Inject()(personService: PersonService,
           journeyCacheService.collectedValues(Seq(UpdateIncome_PayToDateKey, UpdateIncome_IdKey, UpdateIncome_NameKey), Seq(UpdateIncome_DateKey)) flatMap tupled {
             (mandatorySeq, optionalSeq) => {
               val date = optionalSeq.head.map(date => LocalDate.parse(date))
-              EditIncomeForm.bind(mandatorySeq(2), BigDecimal(mandatorySeq.head), date).fold(
-                formWithErrors => {
+              val employerName = mandatorySeq(2)
```
RaminEsfandiari: Like that you have clarified what mandatorySeq(2) is. I think we could do it for pay as well. 
`val payToDate = BigDecimal(mandatorySeq.head)`

app/controllers/IncomeController.scala
```diff
@@ -110,34 +125,49 @@ class IncomeController @Inject()(personService: PersonService,
...
+
+              EditIncomeForm.bind(employerName, BigDecimal(mandatorySeq.head), date).fold(
+                (formWithErrors: Form[EditIncomeForm]) => {
                   val webChat = true
```
RaminEsfandiari: Do we need this line?

app/controllers/IncomeController.scala
```diff
@@ -110,34 +125,49 @@ class IncomeController @Inject()(personService: PersonService,
...
             }
           }
         }
   }
 
+  private def determineEditRedirect(income: EditIncomeForm)(implicit hc: HeaderCarrier): Future[Result] = {
+    for {
+      currentCache <- journeyCacheService.currentCache
+    } yield {
+      val newAmount = income.newAmount.getOrElse("0")
```
RaminEsfandiari: Could maybe inline income.newAmount, and change isCachedIncomeTheSame to take Option[String] since that is what it passes to areEqual. That way in cases of None it would simply return false when comparing to String. My only problem with this is this getOrElse("0") however I can see how further down we need to cache an amount if it's None so we do need the default value of "0". Your thoughts?

app/controllers/IncomeController.scala
```diff
@@ -101,6 +99,23 @@ class IncomeController @Inject()(personService: PersonService,
...
+      implicit request =>
+        ServiceCheckLite.personDetailsCheck {
+          for {
+            cachedData <- journeyCacheService.mandatoryValues(UpdateIncome_NameKey)
+            id <- journeyCacheService.mandatoryValueAsInt(UpdateIncome_IdKey)
+            income <- incomeService.employmentAmount(Nino(user.getNino), id)
+          } yield {
+            val model = SameEstimatedPayViewModel(cachedData(0), income.oldAmount)
+            Ok(views.html.incomes.sameEstimatedPay(model))
+          }
```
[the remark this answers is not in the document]
oc-ste: Agreed

app/controllers/IncomeController.scala
oc-ste: Yeah looks like it turns on webchat when there's something gone wrong when editing pensions

PR #530, opened 2019-04-04
ST-596: Updated underpayment links in all tax-free amount pages

BriWak:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed

There is a corresponding Acceptance test pull request as well


app/controllers/TaxFreeAmountController.scala
```diff
@@ -50,17 +53,21 @@ class TaxFreeAmountController @Inject()(codingComponentService: CodingComponentS
...
       (for {
         codingComponents <- codingComponentService.taxFreeAmountComponents(nino, TaxYear())
         employmentNames <- employmentService.employmentNames(nino, TaxYear())
         companyCarBenefits <- companyCarService.companyCarOnCodingComponents(nino, codingComponents)
+        totalTax <- totalTaxFuture
       } yield {
-        val viewModel = TaxFreeAmountViewModel(codingComponents, employmentNames, companyCarBenefits)
+        totalTax match {
+          case TaiSuccessResponseWithPayload(totalTax: TotalTax) =>
+        val viewModel = TaxFreeAmountViewModel(codingComponents, employmentNames, companyCarBenefits, totalTax)
```
oc-ste: Need to handle the failure case here, the recover won't catch it

app/uk/gov/hmrc/tai/model/CodingComponentPairModel.scala
```diff
@@ -32,12 +33,13 @@ object CodingComponentPairModel {
 
   def apply(codingComponentPair: CodingComponentPair,
             employmentIds: Map[Int, String],
-            companyCarBenefits: Seq[CompanyCarBenefit])
+            companyCarBenefits: Seq[CompanyCarBenefit],
+            totalTax: TotalTax)
            (implicit messages: Messages): CodingComponentPairModel = {
```
oc-ste: The issue here with passing so many things around is that theres no description for this blob of data, if you wrap all of these 4 arguments up in to a case class (not sure what to call it yet) you can give the totalTax, and other params, a default value. This is called the parameter object pattern (https://refactoring.guru/introduce-parameter-object). It should shorten up your tests a lot, especially the view model ones.

app/uk/gov/hmrc/tai/viewModels/TaxFreeAmountViewModel.scala
```diff
@@ -31,7 +32,8 @@ object TaxFreeAmountViewModel extends ViewModelHelper {
 
   def apply(codingComponents: Seq[CodingComponent],
             employmentName: Map[Int, String],
-            companyCarBenefits: Seq[CompanyCarBenefit])
+            companyCarBenefits: Seq[CompanyCarBenefit],
+            totalTax: TotalTax)
            (implicit messages: Messages): TaxFreeAmountViewModel = {
```
oc-ste: Here you'll reuse the new parameter object

app/uk/gov/hmrc/tai/viewModels/TaxSummaryLabel.scala
```diff
@@ -28,29 +31,33 @@ case class HelpLink(value: String, href: String, id: String)
...
 
     TaxSummaryLabel(labelString, labelLink)
   }
 
-  private def createLabelLink(taxComponentType: TaxComponentType)(implicit messages: Messages): Option[HelpLink] = {
+
+  private def createLabelLink(taxComponentType: TaxComponentType, amount: BigDecimal, totalTax: TotalTax)(implicit messages: Messages): Option[HelpLink] = {
+
+    val underpaymentAmount = TaxAmountDueFromUnderpayment.amountDue(amount, totalTax)
+
```
oc-ste: Bug is here, the underpayAmount should only be calculated when we're either a `UnderPaymentFromPreviousYear` or a `EstimatedTaxYouOweThisYear`. Easy fix is just to turn it into a def but I guess there is a missing failing unit test to be placed over this first. To do this, we will need to create a `totalTax` object in the spec that has no `incomeCategories` which will` throw new RuntimeException("Failed to calculate the tax amount due")` and recover with a empty label in the `createLabelLink` function

app/controllers/TaxFreeAmountController.scala
BriWak: Done

app/uk/gov/hmrc/tai/viewModels/TaxSummaryLabel.scala
BriWak: Should be done now if you can re review it

PR #558, opened 2019-04-24
ST-622: Update Content on est pay confirm page to be consistent

BriWak:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed


app/uk/gov/hmrc/tai/viewModels/income/ConfirmAmountEnteredViewModel.scala
```diff
@@ -52,6 +52,23 @@ object ConfirmAmountEnteredViewModel {
     )
   }
 
+  def annualPayCurrentYear(employmentId: Int, employerName: String, currentAmount: Int, estimatedIncome: Int)(implicit messages: Messages): ConfirmAmountEnteredViewModel = {
```
lewey: I think there is an opportunity to clean up this object as there is quite a bit of repeating code. I think at the very least the vals could be in-lined with common ones pushed up. Other options would be to implement a factory or something that resembles "a template method in a super class"

project/FrontendBuild.scala
```diff
@@ -46,7 +46,7 @@ private object AppDependencies {
     def apply() = new TestDependencies {
       override lazy val test = Seq(
         "org.scalatestplus.play" %% "scalatestplus-play" % "2.0.0" % scope,
-        "uk.gov.hmrc"       %%    "hmrctest"   %  "3.5.0-play-25"  % scope,
+        "uk.gov.hmrc"       %%    "hmrctest"   %  "3.8.0-play-25"  % scope,
```
lewey: Unless there is a direct dependency to a new / updated class this should ideally be done in a separate ticket

BriWak: This was just something i missed on a previous PR for dependency updates. I've reverted it now and will raise a separate PR for it

app/uk/gov/hmrc/tai/viewModels/income/ConfirmAmountEnteredViewModel.scala
BriWak: looks like this is done now?

lewey: Yes I managed to take a look, It just needs a review now.

PR #641, opened 2019-08-09
Ddcnls 6539

RobertLaverick:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +114,35 @@ class CompanyBenefitController @Inject()(
...
+                journeyCacheService.cache(getBenefitDecisionKey(Some(name), DecisionChoice), YesIGetThisBenefit) map {
+                  _ =>
+                    Redirect(
+                      controllers.routes.ExternalServiceRedirectController
+                        .auditInvalidateCacheAndRedirectService(TaiConstants.CompanyBenefitsIform)
+                        .url)
+                }
             }
+
+          case Left(e) => throw new NoSuchElementException(e)
```
kurtlogan: Don't throw exceptions.
Prefer return types and logging - here you could have return a status like `InteralServerError` but it really depends on your use case.

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +114,35 @@ class CompanyBenefitController @Inject()(
...
-            journeyCacheService.cache(DecisionChoice, NoIDontGetThisBenefit) map { _ =>
-              Redirect(controllers.benefits.routes.RemoveCompanyBenefitController.stopDate())
-            }
-          case Some(YesIGetThisBenefit) =>
-            journeyCacheService.cache(DecisionChoice, YesIGetThisBenefit) map { _ =>
-              Redirect(
-                controllers.routes.ExternalServiceRedirectController
-                  .auditInvalidateCacheAndRedirectService(TaiConstants.CompanyBenefitsIform)
-                  .url)
+        val benefitType = journeyCacheService.mandatoryJourneyValueAs[String](
```
kurtlogan: Scala should be able to infer the type parameter here

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +114,35 @@ class CompanyBenefitController @Inject()(
...
+          EndCompanyBenefit_BenefitTypeKey,
+          x => x
+        )
+
+        benefitType.flatMap {
+          case Right(name) =>
+            success match {
```
kurtlogan: There is no match case for `None` and are `NoIDontGetThisBenefit` and `YesIGetThisBenefit` the only inhabitants of the type?

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -74,7 +74,8 @@ class CompanyBenefitController @Inject()(
           }
 
           val form = {
-            val decision = currentCache.get(DecisionChoice)
+            val benefitType = currentCache.get(EndCompanyBenefit_BenefitTypeKey)
```
kurtlogan: When is this data set?

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: Redirecting instead

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: Here, yeah. I've still got some of this in the tests as it was complaining about ambiguity.

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: It isn't, and doesn't need to be for the first visit, it is only set on the first submission of the form with these keys.

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +118,41 @@ class CompanyBenefitController @Inject()(
...
+                      .auditInvalidateCacheAndRedirectService(TaiConstants.CompanyBenefitsIform))
+                }
+              case e => {
+                logger.error(s"Data in incorrect format: $e")
+                Future.successful(Redirect(controllers.routes.TaxAccountSummaryController.onPageLoad()))
+              }
             }
+
+          case Left(e) => {
+            logger.error(s"Data not found in cache: $e")
```
kurtlogan: Don't need this logger as it is logged at the connector level

test/controllers/benefits/CompanyBenefitControllerSpec.scala
```diff
@@ -194,6 +208,62 @@ class CompanyBenefitControllerSpec
...
+          RequestBuilder
+            .buildFakeRequestWithAuth("POST")
+            .withFormUrlEncodedBody(DecisionChoice -> NoIDontGetThisBenefit))
+
+        val redirectUrl = redirectLocation(result)
+
+        redirectUrl mustBe Some(controllers.routes.TaxAccountSummaryController.onPageLoad().url)
+
+      }
+      "the form has the value YesIGetThisBenefit and EndCompanyBenefit_BenefitTypeKey is not cached" in {
```
kurtlogan: Add a new line between your tests

test/controllers/benefits/CompanyBenefitControllerSpec.scala
```diff
@@ -194,6 +208,62 @@ class CompanyBenefitControllerSpec
...
+          RequestBuilder
+            .buildFakeRequestWithAuth("POST")
+            .withFormUrlEncodedBody(DecisionChoice -> YesIGetThisBenefit))
+
+        val redirectUrl = redirectLocation(result).getOrElse("")
+
+        redirectUrl mustBe controllers.routes.TaxAccountSummaryController.onPageLoad()
+
+      }
+      "the form has no valid value" in{
```
kurtlogan: New line here please

test/controllers/benefits/CompanyBenefitControllerSpec.scala
```diff
@@ -194,6 +208,62 @@ class CompanyBenefitControllerSpec
...
+        val SUT = createSUT
+        val benefitType = "Telephone"
+        when(
+          journeyCacheService.mandatoryJourneyValueAs[String](Matchers.eq(EndCompanyBenefit_BenefitTypeKey), any())(
```
kurtlogan: Don't need type annotation here
```suggestion
          journeyCacheService.mandatoryJourneyValueAs(Matchers.eq(EndCompanyBenefit_BenefitTypeKey), any())(
```
This is the same for the rest of the file

kurtlogan: You can replace the `Matchers.eq` with named imports to make it a little more concise.
If you import `import org.mockito.Matchers.{eq=>eqTo}` then usage would be 
```suggestion
          journeyCacheService.mandatoryJourneyValueAs[String](eqTo(EndCompanyBenefit_BenefitTypeKey), any())(
```
As this is used a lot in this file it would make it much easier to read the code

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: Removed

test/controllers/benefits/CompanyBenefitControllerSpec.scala
RobertLaverick: Added

test/controllers/benefits/CompanyBenefitControllerSpec.scala
RobertLaverick: added

test/controllers/benefits/CompanyBenefitControllerSpec.scala
RobertLaverick: Added and replaced all instances in file

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -45,6 +47,8 @@ class CompanyBenefitController @Inject()(
   override implicit val partialRetriever: FormPartialRetriever)
     extends TaiBaseController with JourneyCacheConstants with UpdateOrRemoveCompanyBenefitDecisionConstants {
 
+  private val logger = Logger(this.getClass)
+
```
oc-ste: Created a new logger but it's not being used, this needs to be added to conf.xml and logback.xml, try not to add genuine exceptions into here, as this logger is probably not tracked by OS team

RobertLaverick: I mean, it is being used (admittedly only in one case) - Line:142.

But yeah, I can add it to those files.

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +118,40 @@ class CompanyBenefitController @Inject()(
...
-          case Some(YesIGetThisBenefit) =>
-            journeyCacheService.cache(DecisionChoice, YesIGetThisBenefit) map { _ =>
-              Redirect(
-                controllers.routes.ExternalServiceRedirectController
-                  .auditInvalidateCacheAndRedirectService(TaiConstants.CompanyBenefitsIform)
-                  .url)
+        val benefitType = journeyCacheService.mandatoryJourneyValueAs(
+          EndCompanyBenefit_BenefitTypeKey,
+          x => x
+        )
```
oc-ste: x => x doesn't do anything here, you can use 

```suggestion
        ) journeyCacheService.mandatoryJourneyValue(EndCompanyBenefit_BenefitTypeKey)
```

app/controllers/benefits/CompanyBenefitController.scala
oc-ste: It's getting logged on Line:142 but you'll never be able to see it until its added to the logger configs

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +118,40 @@ class CompanyBenefitController @Inject()(
...
+          case Left(e) => {
+            Future.successful(Redirect(controllers.routes.TaxAccountSummaryController.onPageLoad()))
+          }
         }
       }
     )
   }
+
+  def getBenefitDecisionKey(benefitType: Option[String], decision: String): String =
+    s"$benefitType $decision"
```
oc-ste: Do you want to interpolate over an Option here? It'll return something like "None Decision" or "Some(benefitType) decision"

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +118,40 @@ class CompanyBenefitController @Inject()(
...
-              Redirect(controllers.benefits.routes.RemoveCompanyBenefitController.stopDate())
-            }
-          case Some(YesIGetThisBenefit) =>
-            journeyCacheService.cache(DecisionChoice, YesIGetThisBenefit) map { _ =>
-              Redirect(
-                controllers.routes.ExternalServiceRedirectController
-                  .auditInvalidateCacheAndRedirectService(TaiConstants.CompanyBenefitsIform)
-                  .url)
+        val benefitType = journeyCacheService.mandatoryJourneyValueAs(
+          EndCompanyBenefit_BenefitTypeKey,
```
oc-ste: I think there is a bug somewhere in this file related to this key. The function `redirectCompanyBenefitSelection` caches the `EndCompanyBenefit_BenefitTypeKey  -> benefitType.toString` but it's being used here against NoIDontGetThisBenefit and YesIGetThisBenefit. I suggest adding in a new key or figuring out if it's ever actually being used in `redirectCompanyBenefitSelection`

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +118,40 @@ class CompanyBenefitController @Inject()(
...
+                    Redirect(controllers.benefits.routes.RemoveCompanyBenefitController.stopDate())
+                  }
+              case Some(YesIGetThisBenefit) =>
+                journeyCacheService.cache(getBenefitDecisionKey(Some(name), DecisionChoice), YesIGetThisBenefit).map {
+                  _ =>
+                    Redirect(controllers.routes.ExternalServiceRedirectController
+                      .auditInvalidateCacheAndRedirectService(TaiConstants.CompanyBenefitsIform))
+                }
+              case e => {
+                logger.error(s"Data in incorrect format: $e")
```
oc-ste: Can you make this error message less generic?

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: No, Ideally I do want to have just "Telephone NoIDontGetThisBenefit", so would you recommend I fold benefit type so I get "Telephone NoIDontGetThisBenefit" if there is a value or "NoIDontGetThisBenefit" if there is no benefitType value?

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +118,40 @@ class CompanyBenefitController @Inject()(
...
+                Future.successful(Redirect(controllers.routes.TaxAccountSummaryController.onPageLoad()))
+              }
             }
+
+          case Left(e) => {
+            Future.successful(Redirect(controllers.routes.TaxAccountSummaryController.onPageLoad()))
```
oc-ste: `e` is not used here, replace with wildcard `_` or use it in a logger

test/controllers/benefits/CompanyBenefitControllerSpec.scala
```diff
@@ -25,14 +27,15 @@ import org.jsoup.Jsoup
...
+import org.mockito.Matchers.{eq=>eqTo}
 import org.scalatest.BeforeAndAfterEach
 import org.scalatest.mockito.MockitoSugar
 import org.scalatestplus.play.PlaySpec
 import play.api.data.Form
 import play.api.i18n.{I18nSupport, Messages, MessagesApi}
 import play.api.test.Helpers.{contentAsString, status, _}
 import uk.gov.hmrc.domain.{Generator, Nino}
-import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.http.{HeaderCarrier, NotFoundException}
```
oc-ste: NotFoundException has been added, but it's not used, can you optimise the imports?

test/controllers/benefits/CompanyBenefitControllerSpec.scala
```diff
@@ -194,6 +210,64 @@ class CompanyBenefitControllerSpec
...
+            .buildFakeRequestWithAuth("POST")
+            .withFormUrlEncodedBody(DecisionChoice -> Random.alphanumeric.take(10).mkString))
+
+        status(result) mustBe SEE_OTHER
+
+        val redirectUrl = redirectLocation(result).getOrElse("")
+
+        redirectUrl mustBe controllers.routes.TaxAccountSummaryController.onPageLoad().url
+      }
+    }
```
oc-ste: Sounds overkill but there should be a test for the actual bug of where there is a good (Right result from mandatoryJourneyValueAs) but it does not match with the user's current journey

test/controllers/benefits/CompanyBenefitControllerSpec.scala
```diff
@@ -194,6 +210,64 @@ class CompanyBenefitControllerSpec
...
+      }
+
+      "the form has no valid value" in{
+        val SUT = createSUT
+        val benefitType = "Telephone"
+        when(
+          journeyCacheService.mandatoryJourneyValueAs[String](eqTo(EndCompanyBenefit_BenefitTypeKey), any())(
+            any()))
+          .thenReturn(Future.successful(Right(benefitType)))
+
```
oc-ste: You can deduplicate this block into a function within this `should` scope

`val benefitType = "Telephone"
        when(
          journeyCacheService.mandatoryJourneyValueAs[String](eqTo(EndCompanyBenefit_BenefitTypeKey), any())(
            any()))
          .thenReturn(Future.successful(Right(benefitType)))`

test/controllers/benefits/CompanyBenefitControllerSpec.scala
```diff
@@ -176,10 +186,16 @@ class CompanyBenefitControllerSpec
     }
 
     "redirect to the appropriate IFORM update page" when {
-      "the form has the value yesIGetThisBenefit" in {
+      "the form has the value yesIGetThisBenefit and EndCompanyBenefit_BenefitTypeKey is cached" in {
 
         val SUT = createSUT
 
+        val benefitType = "Telephone"
```
oc-ste: To be explicit about it being a benefit you could Telephone type in `TaxComponentType`

test/controllers/benefits/CompanyBenefitControllerSpec.scala
RobertLaverick: As in, returns Yes when it should be No? Or do you mean it returns Apples when it should be Yes or No?

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -113,20 +118,40 @@ class CompanyBenefitController @Inject()(
...
+                  }
+              case Some(YesIGetThisBenefit) =>
+                journeyCacheService.cache(getBenefitDecisionKey(Some(name), DecisionChoice), YesIGetThisBenefit).map {
+                  _ =>
+                    Redirect(controllers.routes.ExternalServiceRedirectController
+                      .auditInvalidateCacheAndRedirectService(TaiConstants.CompanyBenefitsIform))
+                }
+              case e => {
+                logger.error(s"Data in incorrect format: $e")
+                Future.successful(Redirect(controllers.routes.TaxAccountSummaryController.onPageLoad()))
```
oc-ste: You can deduplicate this, extract it to a val

app/controllers/benefits/CompanyBenefitController.scala
oc-ste: Mapping over benefitType first and ensure you always pass in a string into `getBenefitDecisionKey` is probably the way to go as you'd never call this function on a None

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: Right, the match is being done on success rather than name - and success is the DecisionChoice.

I'll refactor the flatmap to make the storage only happen before the match statement to make it clearer.

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: I remember trying this, and it leads to type ambiguity (if I understand the error correctly).

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: this case is an invalid decision being provided.

oc-ste: It's not about the case but about the actual message being logged in, it's quite generic when it could be specific.

test/controllers/benefits/CompanyBenefitControllerSpec.scala
oc-ste: As in, the cache is set up to return one benefit and when you start the journey on another benefit it's not prepopulated - basically a test for the scenario described in the ticket. If it's not possible here then it needs to be done in an acceptance test.

app/controllers/benefits/CompanyBenefitController.scala
RobertLaverick: oh, no, I get that - I've improved this (commit pending), I was just stating that it isn't a match against the benefitType.

test/controllers/benefits/CompanyBenefitControllerSpec.scala
RobertLaverick: I've done this now for ensuring it exists, and ensuring it is clear (complete with the use of Telephone.name)

test/controllers/benefits/CompanyBenefitControllerSpec.scala
RobertLaverick: I'm going to handle this with an acceptance test.

app/controllers/benefits/CompanyBenefitController.scala
oc-ste: Try changing it to (note removing the As)

>     val benefitType = journeyCacheService.mandatoryJourneyValue(EndCompanyBenefit_BenefitTypeKey)

app/controllers/benefits/CompanyBenefitController.scala
oc-ste: getBenefitDecisionKey now has different behaviour in the decision function. It's obtaining the cache values on `$suffix` which is never cached, so you will always get a None back when calling `journeycacheService.getCurrentValue("suffix is never cached")`

It's misleading and can make it difficult to read and debug at a later date. I suggest making the `getBenefitDecisionKey` return an Option[String], this will add a bit of code complexity as you have to match on this option but it means the behaviour hasn't changed pre-refactor and you know you're trying to get a None from `journeycacheService.getCurrentValue` when you initially land on this page. 

To make it easier to read and to prevent someone accidentally changing the cached/extracted key, I would pull the caching logic to its own file.

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -104,6 +101,22 @@ class CompanyBenefitController @Inject()(
...
+      case NoIDontGetThisBenefit => {
+        Redirect(controllers.benefits.routes.RemoveCompanyBenefitController.stopDate())
+      }
+      case YesIGetThisBenefit => {
+        Redirect(
+          controllers.routes.ExternalServiceRedirectController
+            .auditInvalidateCacheAndRedirectService(TaiConstants.CompanyBenefitsIform))
+      }
+      case e => {
+        logger.error(s"Bad Option provided in submitDecision form: $e")
```
oc-ste: Should be NonFatal(e)

app/uk/gov/hmrc/tai/DecisionCacheWrapper.scala
```diff
@@ -0,0 +1,86 @@
...
+      }
+      case Left(_) => {
+        logger.error(s"Unable to find $EndCompanyBenefit_BenefitTypeKey when retrieving decision")
+        Future.successful(journeyStartRedirection)
+      }
+    }
+  }
+
+  def getBenefitDecisionKey(benefitType: Option[String]): Option[String] =
+    benefitType.map(x => s"$x $DecisionChoice")
```
oc-ste: Can this be private?

test/controllers/benefits/CompanyBenefitControllerSpec.scala
```diff
@@ -124,9 +126,11 @@ class CompanyBenefitControllerSpec
         when(journeyCacheService.currentCache(any())).thenReturn(Future.successful(cache))
         when(employmentService.employment(any(), any())(any())).thenReturn(Future.successful(Some(employment)))
         when(journeyCacheService.cache(any())(any())).thenReturn(Future.successful(Map("" -> "")))
+        when(journeyCacheService.mandatoryJourneyValue(any())(any())).thenReturn(Future.successful(Right(benefitType)))
+        when(journeyCacheService.currentValue(any())(any())).thenReturn(Future.successful(Some(s"$benefitType $YesIGetThisBenefit")))
 
         val expectedForm: Form[Option[String]] =
-          UpdateOrRemoveCompanyBenefitDecisionForm.form.fill(Some(YesIGetThisBenefit))
+          UpdateOrRemoveCompanyBenefitDecisionForm.form.fill(Some(s"$benefitType $YesIGetThisBenefit"))
```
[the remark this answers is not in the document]
RobertLaverick: No, that should not be there. I'll get rid of that sharpish

PR #647, opened 2019-08-30
DDCNOS-4268 | returning 502 when RTI is unavailable

shikkhasingh:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/YourIncomeCalculationController.scala
+++ b/app/controllers/YourIncomeCalculationController.scala
@@ -70,7 +70,7 @@ class YourIncomeCalculationController @Inject()(
       employmentDetails    <- employmentFuture
     } yield {
       (taxCodeIncomeDetails, employmentDetails) match {
-        case (TaiSuccessResponseWithPayload(taxCodeIncomes: Seq[TaxCodeIncome]), Some(employment)) =>
+        case (TaiSuccessResponseWithPayload(taxCodeIncomes: Seq[TaxCodeIncome]), Some(employment)) =>{
           val paymentDetails = paymentsService.filterDuplicates(employment)
 
           val model = YourIncomeCalculationViewModel(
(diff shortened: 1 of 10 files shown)


app/controllers/YourIncomeCalculationController.scala
```diff
@@ -122,11 +122,13 @@ class YourIncomeCalculationController @Inject()(
       employment <- employmentService.employments(nino, year)
     } yield {
       val historicIncomeCalculationViewModel = HistoricIncomeCalculationViewModel(employment, empId, year)
-      if (printPage) {
-        Ok(views.html.print.historicIncomeCalculation(historicIncomeCalculationViewModel))
-      } else {
-        Ok(views.html.incomes.historicIncomeCalculation(historicIncomeCalculationViewModel))
+      val printablePage = views.html.print.historicIncomeCalculation(historicIncomeCalculationViewModel)
```
lewey: The two view refs should be lazy

app/controllers/YourIncomeCalculationController.scala
```diff
@@ -122,11 +122,13 @@ class YourIncomeCalculationController @Inject()(
...
+      val incomePage = views.html.incomes.historicIncomeCalculation(historicIncomeCalculationViewModel)
+      (printPage, historicIncomeCalculationViewModel.realTimeStatus.toString) match {
```
lewey: There could be a potential to create another view to cater for the error path. That way the view would not need to be bothered about checking for the RTI status

shikkhasingh: Removed the checking for RTI status "TemporarilyUnavailable" in html file and using the serviceUnvailable view if RTI is down.

PR #651, opened 2019-09-10
DDCNOS-4271 | putting check for empty list

shikkhasingh:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/PayeControllerHistoric.scala
+++ b/app/controllers/PayeControllerHistoric.scala
@@ -16,19 +16,18 @@
 
 package controllers
 
-import javax.inject.Inject
 import controllers.actions.ValidatePerson
 import controllers.auth.{AuthAction, AuthenticatedRequest}
+import javax.inject.Inject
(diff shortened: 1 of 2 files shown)


app/controllers/PayeControllerHistoric.scala
```diff
@@ -82,7 +82,7 @@ class PayeControllerHistoric @Inject()(
   } recoverWith hodStatusRedirect
 
   private def isRtiUnavailable(employments: Seq[Employment]): Boolean =
-    (employments.head.annualAccounts.head.realTimeStatus.toString == "TemporarilyUnavailable")
+    (!employments.isEmpty && !employments.head.annualAccounts.isEmpty && (employments.head.annualAccounts.head.realTimeStatus.toString == "TemporarilyUnavailable"))
```
liampan: we can un-negate these if we use `.nonEmpty`

PR #655, opened 2019-10-02
Ddcnls 6754: Split tax codes to separate pages

oc-ste:
- [x] Unit tests passing
- [x] Coverage reached
- [x] Reviewed


test/uk/gov/hmrc/tai/viewModels/IncomeSourceViewModelSpec.scala
```diff
@@ -0,0 +1,280 @@
...
+import uk.gov.hmrc.tai.model.domain.{JobSeekerAllowanceIncome, OtherIncome, PensionIncome, TaxedIncome}
+import utils.TaxAccountSummaryTestData
+import uk.gov.hmrc.tai.util.constants.TaiConstants.encodedMinusSign
+
+class IncomeSourceViewModelSpec
+    extends PlaySpec with FakeTaiPlayApplication with I18nSupport with TaxAccountSummaryTestData {
+
+  implicit val messagesApi: MessagesApi = app.injector.instanceOf[MessagesApi]
+
+  "IncomeSourceViewModel apply method (two parameter version)" must {
```
kurtlogan: Remove two parameter version as it is no longer relevant 
```suggestion
  "IncomeSourceViewModel apply method" must {
```

app/uk/gov/hmrc/tai/viewModels/IncomeSourceViewModel.scala
```diff
@@ -35,11 +35,12 @@ case class IncomeSourceViewModel(
   displayEndDate: Boolean,
   detailsLinkLabel: String,
   detailsLinkUrl: String,
+  taxCodeUrl: String = "",
```
kurtlogan: Is a string the best data type to represent this. I would prefer a Call over a string and if it has the option of not being required maybe a `Option[Call]` would be suitable

PR #666, opened 2019-11-20
css updated to reduce heading margin

Muhammad-Ali-Ahmad:
- [ yes ] Unit tests passing
- [ yes ] Coverage reached
- [ yes ] Acceptance tests passing
- [ yes ] Reviewed

### DIFF
--- a/app/assets/sass/tai19032019.scss
+++ b/app/assets/sass/tai19032019.scss
@@ -168,7 +168,3 @@ hr {
   right: 0;
   z-index: 9999;
 }
-
-.incomeSourceSummary .heading-48 {
-    margin-bottom: .625em;
-}
--- a/app/assets/sass/tai19032019.scss
+++ b/app/assets/sass/tai19032019.scss
@@ -168,3 +168,7 @@ hr {
   right: 0;
   z-index: 9999;
 }
+
+.incomeSourceSummary .heading-48 {
+    margin-bottom: .625em;
+}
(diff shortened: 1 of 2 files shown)


app/assets/sass/tai19032019.scss
```diff
@@ -167,4 +167,8 @@ hr {
   bottom: 0;
   right: 0;
   z-index: 9999;
-}
\ No newline at end of file
+}
+
+.incomeSourceSummary .heading-48 {
+    margin-bottom: .625em;
```
timsb: What font-size does the resolve to? Wouldn't it be better to change the class used to another heading size?

app/views/IncomeSourceSummary.scala.html
```diff
@@ -76,7 +76,7 @@
     title = title,
     authedUser = Some(user),
     pageHeader = Some(header),
-    articleClasses = None,
+    articleClasses = Some("incomeSourceSummary"),
```
timsb: Do you still need this class?

Muhammad-Ali-Ahmad: I won't if Rachel approves the design that only requires class update

app/assets/sass/tai19032019.scss
Muhammad-Ali-Ahmad: Updated as per suggestion

app/views/IncomeSourceSummary.scala.html
Muhammad-Ali-Ahmad: updated as per suggestion

PR #688, opened 2020-02-07
DDCNL-217 | Revise Page not found

jpatton638:
- [X] Unit tests passing
- [x] Coverage reached
- [x] Reviewed

### DIFF
--- a/app/views/error_template_noauth.scala.html
+++ b/app/views/error_template_noauth.scala.html
@@ -24,7 +24,7 @@
     <section class="section">
         <h1 class="h1-heading">@heading</h1>
         <div class="grid grid-2-3">
-            <p id="paraOne">@Html(message1)</p>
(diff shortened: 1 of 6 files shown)


app/views/error_template_noauth.scala.html
```diff
@@ -25,11 +25,8 @@
         <h1 class="h1-heading">@heading</h1>
         <div class="grid grid-2-3">
             <p id="paraOne">@Html(message1)</p>
-            @message2 match {
-                case Some(message2) => {
-                    <p id="paraTwo">@Html(message2)</p>
```
oc-ste: Lost paraTwo here, intentional?

RobertLaverick: If it is, can you also drop paraOne

jpatton638: :+1:

PR #700, opened 2020-03-11
DDCNL-1526 | Add support for CY-5

jpatton638:
- [X] Unit tests passing
- [X] Coverage reached
- [ ] Reviewed

### DIFF
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -52,7 +52,7 @@ ur-banner.url = "https://www.surveygizmo.com/s3/4856606/HMRC-Letter-Study-7"
 
 tai.cyPlusOne.enabled = true
 tai.cy3.enabled = false
-tai.numberOfPreviousYearsToShow = 4
+tai.numberOfPreviousYearsToShow = 5
 tai.feature.welshLanguage.enabled = true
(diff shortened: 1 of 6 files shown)


conf/application.conf
```diff
@@ -52,7 +52,7 @@ ur-banner.url = "https://www.surveygizmo.com/s3/4856606/HMRC-Letter-Study-7"
 
 tai.cyPlusOne.enabled = true
 tai.cy3.enabled = false
-tai.numberOfPreviousYearsToShow = 4
+tai.numberOfPreviousYearsToShow = 5
```
oc-ste: Think there's a ticket somewhere mentioning this toggle is never used so it doesn't need to be an application config thing, as it's not a config it's just a val.

jpatton638: Do you want me to undo this? The val in the ApplicationConfig file will default it to 5 years anyway. I tested this change locally and it brought back CY-5 so it does appear to work. Thoughts?

oc-ste: Up to you if you want to delete it or just leave it for the next ticket. Figured it would be a good time to delete it now as you've already had a look into the usage of the toggle.

PR #721, opened 2020-06-09
DEP-276

Deep547:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/views/main.scala.html
```diff
@@ -86,6 +89,38 @@
...
+                "mdtpdfSessionID" : "@{nuanceData.nuanceSessionId}",
+                "mdtpSessionID" : "@{nuanceData.mdtpSessionID}",
+                "deviceID" : "@{nuanceData.deviceID}"
+            }
+        </script>
+        <script id="webchat-tag" type="text/javascript" src=@nuanceUrl></script>
+    }
+}
+
+
```
Burdis, line 118: Remove whitespace

app/models/EncryptedNuanceData.scala
```diff
@@ -0,0 +1,59 @@
+/*
```
tonymyers, line 1: The Encryption model is in the incorrect place. It should be in uk.gov.hmrc.tai.models folder

app/services/NuanceEncryptionService.scala
```diff
@@ -0,0 +1,61 @@
+/*
```
tonymyers, line 1: The Nuance Encryption service is in the incorrect place. It should be in uk.gov.hmrc.tai.services folder

PR #736, opened 2020-09-14
DDCNL-2688 | Remove use of deprecated SessionKeys attributes

jpatton638:
- [x] Unit tests passing
- [x] Coverage reached
- [x] Reviewed

### DIFF
--- a/app/views/main.scala.html
+++ b/app/views/main.scala.html
@@ -55,11 +55,11 @@
 @isUserResearchBannerHidden = @{request.cookies.exists((x: Cookie) => x.name == "mdtpurr")}
 
 @authProvider = @{
-    request.session.get(TaiConstants.AuthProvider)
+    request.session.get(TaiConstants.ViewAuthProvider)
 }
 
 @isGG = @{
-    authProvider.contains(TaiConstants.AuthProviderGG)
+    authProvider.contains(TaiConstants.ViewAuthProviderGG)
 }
 
 @isSa = @{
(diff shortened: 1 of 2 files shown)


oc-ste (approved):

app/views/main.scala.html
```diff
@@ -55,11 +55,11 @@
 @isUserResearchBannerHidden = @{request.cookies.exists((x: Cookie) => x.name == "mdtpurr")}
 
 @authProvider = @{
-    request.session.get(SessionKeys.authProvider)
+    request.session.get(TaiConstants.AuthProvider)
```
oc-ste, line 58: @vimalsuresh need to check we can log in via verify and gg fine and that the sign out redirects you to ida/signout and gg/signout

app/views/main.scala.html
```diff
@@ -69,7 +69,7 @@
 }
 
 @isVerify = @{
-    authProvider.contains("IDA")
+    authProvider.contains(TaiConstants.AuthProviderVerify)
```
oc-ste, line 72: @vimalsuresh need to check the correct menu bar is shown but I believe this is under test via ATs

oc-ste (approved):

oc-ste (approved):

app/views/main.scala.html
vimalsuresh: Tested in local , looks good

app/views/main.scala.html
vimalsuresh: Have AT's to test menu links, looks good

PR #762, opened 2021-02-12
Ddcnl 481

jpatton638:
- [X] Unit tests passing
- [X] Coverage reached
- [X] Acceptance tests passing
- [ ] Reviewed


app/uk/gov/hmrc/tai/service/AuditService.scala
```diff
@@ -86,9 +89,9 @@ class AuditService @Inject()(
     endDate: String,
     fuelBenefitDate: Option[String],
     isSuccessful: Boolean,
-    path: String)(implicit hc: HeaderCarrier): Future[AuditResult] = {
+    path: String)(implicit hc: HeaderCarrier, request: Request[_]): Future[AuditResult] = {
     val details = Map(
-      "authProviderId" -> authProviderId(hc),
+      "authProviderId" -> authProviderId,
```
struong, line 94: I think you can just delete this, we will probably remove this audit event when we get to the ticket https://jira.tools.tax.service.gov.uk/browse/DDCNL-3096 

as it only audits on the landing page, so we won't see it if a user comes in from mobile.

struong (approved):

PR #782, opened 2021-04-15
DEP-1438: Integrate Webchat library, including Dependency Injection of views as needed.

jaynabonne:
- [x] Unit tests passing
- [x] Coverage reached
- [x] Reviewed


app/controllers/ErrorPagesHandler.scala
```diff
@@ -29,19 +29,23 @@ import uk.gov.hmrc.tai.model.domain.Employment
...
 import uk.gov.hmrc.tai.util.constants.TaiConstants._
 import uk.gov.hmrc.urls.Link
+import views.html.{error_no_primary, error_template_noauth}
 
 import scala.concurrent.Future
 
 trait ErrorPagesHandler {
 
+  val error_template_noauth: error_template_noauth
+  val error_no_primary: error_no_primary
```
struong, line 39: If you make ErrorPagesHandler a class and inject it in to the controllers that use it, I think it'll tidy a lot of this code up as we've ended up (due to `TaiBaseController` extending `ErrorPagesHandler`)  everything being an `ErrorPagesHandler` when it really does not need to be.

struong (changes requested):
`HistoricIncomeCalculationView.scala.html` 
Has been renamed but main has not been injected

struong (approved):

struong (approved):

PR #792, opened 2021-05-26
Ddcnl 3945

Muhammad-Ali-Ahmad:


app/views/session_expired.scala.html
```diff
@@ -0,0 +1,39 @@
+@*
```
struong, line 1: Rename to SessionExpiredView

app/views/unauthenticated_main.scala.html
```diff
@@ -0,0 +1,109 @@
...
+)
+
+@(title: String,
+  sidebarLinks: Option[Html] = None,
+  sidebarClass: Option[String] = None,
+  articleClasses: Option[String] = Some("full-width"),
+  includeGridWrapper: Boolean = false,
+  script: Option[Html] = None,
+  pageHeader: Option[Html] = None,
+  displayTimeoutDialog: Boolean = true
```
struong, line 46: Is this used in this file?

app/views/main.scala.html
```diff
@@ -50,7 +50,8 @@
   optimizelyAudience: Option[String] = None,
   pageHeader: Option[Html] = None,
   script: Option[Html] = None,
-  hidePtaAccountNav: Boolean = false
+  hidePtaAccountNav: Boolean = false,
+  displayTimeoutDialog: Boolean = true
```
struong, line 54: Can't see where this is being set too false, do you need it?

app/views/session_expired.scala.html
Muhammad-Ali-Ahmad: done

app/views/main.scala.html
Muhammad-Ali-Ahmad: yes its not used as only unauthenticated main is used for session expired page where we dont want to display timeout dialog so i've removed it

app/views/unauthenticated_main.scala.html
Muhammad-Ali-Ahmad: removed

app/uk/gov/hmrc/tai/config/ApplicationConfig.scala
```diff
@@ -109,6 +109,9 @@ class ApplicationConfig @Inject()(
 
   lazy val citizenAuthFrontendSignOutUrl: String = citizenAuthHost + "/ida/signout"
 
+  lazy val sessionTimeoutInSeconds: Int = 900
+  lazy val sessionCountdownInSeconds: Int = 120
```
struong, line 113: Can we read these from a config? We sometimes want to change it on different environments to help testing

Muhammad-Ali-Ahmad: done

PR #815, opened 2021-09-03
DDCNL-3582: Added months in Welsh

Christian-QA:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed


Sachin300793 (approved):

app/uk/gov/hmrc/tai/model/YearAndMonth.scala
```diff
@@ -16,24 +16,45 @@
...
 
-  def formatYearAndMonth: String =
-    yearAndMonth.toString(dateFormatter)
+  def formatYearAndMonth(lang: Lang): String =
+    if (lang == Lang("cy")) {
+      val month = MonthNamesInWelsh(yearAndMonth.getMonthOfYear)
+      month + " " + yearAndMonth.getYear
+    } else {
+      yearAndMonth.toString(dateFormatter)
+    }
```
struong, line 36: Duplicated code with `timeout`

Christian-QA: Done

app/uk/gov/hmrc/tai/model/YearAndMonth.scala
```diff
@@ -58,9 +74,14 @@ object YearAndMonth {
     dateTypeList.sortWith((x, y) => x.yearAndMonth.isBefore(y.yearAndMonth))
   }
 
-  private def firstClaimDate(appConfig: ApplicationConfig): YearMonth =
+  def firstClaimDate(appConfig: ApplicationConfig): YearMonth =
```
struong, line 77: Can this still be private?

Christian-QA: I tried for a while, but had to make it public so it can be called in the views with a YearMonth rather than an AppConfig in the params for formattedDate.
The only other way I could figure out that wasn't overcomplicated had an Option[YearMonth] in both and a conditional to check whether it was Some(value) or None, the None case using the firstClaimDate instead.

struong (approved):

PR #819, opened 2021-09-24
DDCNL-4634: Fixes for TAI unknown failing accessibility tests in jenkins 

Muhammad-Ali-Ahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/assets/javascripts/char-count.js
```diff
@@ -166,7 +172,10 @@
     // Check for existing info count message
     var countMessage = document.getElementById(elementId + '-info')
     // If there is no existing info count message we add one right after the field
-    if (elementId && !countMessage) {
```
oscarduignan, line 169: @Muhammad-Ali-Ahmad I'm rereading this now and I think earlier I missed the fact that there was already a guard on this one. My apologies, I did a search for "insertAdjacentHTML" and didn't look closely enough at the context. The original code on this part should be ok to revert to

app/assets/javascripts/char-count.js
```diff
@@ -81,21 +81,27 @@
...
-          var countHighlight = document.getElementById(elementId + '-hl')
-          // countHighlight.style.height = CharCount.prototype.getHeight(countElement) + 'px'
-          // countHighlight.style.height = countElement.getBoundingClientRect().height + 'px'
-          countHighlight.style.height = countElement.offsetHeight + 'px'
-          countHighlight.style.width = countElement.offsetWidth + 'px'
 
-          // We have to disable resize on highlighted components to avoid the async scroll and boundaries
-          countElement.style.resize = 'none'
+          if(document.getElementById(elementId + '-hl') == null) {
+            wrapper.insertAdjacentHTML('afterbegin', '<div id="' + elementId + '-hl" class="form-control ' + countHighlightClass + '" aria-hidden="true" role="presentation"></div>')
```
oscarduignan, line 87: @Muhammad-Ali-Ahmad I think it might just be line 87 we need to wrap in a conditional, though I'm not sure from reading the other code that will be skipped if that code is actually doing anything useful in the context of an accessibility test and that's the only time this will trigger so might be fine as it is 👍

oscarduignan (approved):

oliwiaKot (approved):

russellthorn (approved):

PR #836, opened 2022-01-13
Ddcnl 4992

Muhammad-Ali-Ahmad:


app/assets/stylesheets/tai.scss
```diff
@@ -202,3 +202,26 @@ hr {
...
     color: #005ea5;
 }
+
+#headingInLabel {
+    display: table;
+    table-layout: fixed;
+    width: 100%;
+
+    .form-label{
+        font-family: "nta",Arial,sans-serif;
```
pascal22p, line 212: How come we specify the font name in tai? Is it not defined globally? Plus arial is purely microsoft as far as I know.

pascal22p (approved):

app/assets/stylesheets/tai.scss
Muhammad-Ali-Ahmad: yeah you are right we dont need that. i had copied h1 styles as they were. I have removed it

Luka-Abey (approved):

PR #854, opened 2022-03-09
PAYEE-119 - add print button and update css

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/views/includes/ptaHeader.scala.html
```diff
@@ -45,6 +45,19 @@
 
     </div>
 }
+
+
+<div class="grid-row">
+    <div class="column-full">
+       <div class="print-banner">
+        <div class="hmrc-logo"><img src="/check-income-tax/assets/images/HMRC-logo.png" alt='HM Revenue & Customs'></div>
```
sergi-hmrc, line 53: should the message come from a label?

app/assets/stylesheets/modules/_print.scss
```diff
@@ -442,4 +442,72 @@ tr.print-pension-contributions-data td {
...
+  }
+
+
+body.printable-page .cbanner-govuk-cookie-banner {
+  display: none;
+  visibility: hidden;
+}
+
+
+  body.printable-page a[href],
```
sergi-hmrc, line 498: should we define the styles hierarchically?

app/views/includes/ptaHeader.scala.html
masahmad: do you mean the alt attribute  - if so  then I think we should leave as is - because its redundant in this case (ie because its for printing only and not screen or screen reader use

app/assets/stylesheets/modules/_print.scss
masahmad: I wouldn't worry too much about it - if the hierarchy fails then we would also notice on the page with styles overriding incorrectly - but yeh scss you can nest things

sergi-hmrc (approved):

PR #860, opened 2022-03-24
Paye 183

vivekhmrc:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/WhatDoYouWantToDoController.scala
+++ b/app/controllers/WhatDoYouWantToDoController.scala
@@ -130,7 +130,7 @@ class WhatDoYouWantToDoController @Inject()(
         } yield Ok(whatDoYouWantToDoTileView(WhatDoYouWantToDoForm.createForm, model, applicationConfig))
       case Left(taxCodeError) =>
         Future.successful(
-          errorPagesHandler.error4xxPageWithLink(taxCodeError.errorMessage.getOrElse("Tax Code not Found")))
+          BadRequest(errorPagesHandler.error4xxPageWithLink(taxCodeError.errorMessage.getOrElse("Tax Code not Found"))))
     }
   }
 
--- a/app/controllers/WhatDoYouWantToDoController.scala
+++ b/app/controllers/WhatDoYouWantToDoController.scala
@@ -130,7 +130,7 @@ class WhatDoYouWantToDoController @Inject()(
         } yield Ok(whatDoYouWantToDoTileView(WhatDoYouWantToDoForm.createForm, model, applicationConfig))
       case Left(taxCodeError) =>
         Future.successful(
-          BadRequest(errorPagesHandler.error4xxPageWithLink(taxCodeError.errorMessage.getOrElse("Tax Code not Found"))))
+          errorPagesHandler.error4xxPageWithLink(taxCodeError.errorMessage.getOrElse("Tax Code not Found")))
     }
   }
 
--- a/app/controllers/WhatDoYouWantToDoController.scala
+++ b/app/controllers/WhatDoYouWantToDoController.scala
@@ -18,25 +18,23 @@ package controllers
 
 import controllers.actions.ValidatePerson
 import controllers.auth.{AuthAction, AuthedUser}
-
-import javax.inject.Inject
 import play.api.Logging
 import play.api.mvc._
 import uk.gov.hmrc.domain.Nino
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.audit.http.connector.{AuditConnector, AuditResult}
-
 import uk.gov.hmrc.renderer.TemplateRenderer
 import uk.gov.hmrc.tai.config.ApplicationConfig
 import uk.gov.hmrc.tai.connectors.responses.{TaiNotFoundResponse, TaiResponse, TaiSuccessResponseWithPayload}
 import uk.gov.hmrc.tai.forms.WhatDoYouWantToDoForm
 import uk.gov.hmrc.tai.model.TaxYear
-import uk.gov.hmrc.tai.model.domain.{Employment, HasTaxCodeChanged}
 import uk.gov.hmrc.tai.model.domain.income.TaxCodeIncome
+import uk.gov.hmrc.tai.model.domain.{Employment, HasTaxCodeChanged}
 import uk.gov.hmrc.tai.service._
 import uk.gov.hmrc.tai.viewModels.WhatDoYouWantToDoViewModel
 import views.html.WhatDoYouWantToDoTileView
 
+import javax.inject.Inject
 import scala.concurrent.{ExecutionContext, Future}
 
 class WhatDoYouWantToDoController @Inject()(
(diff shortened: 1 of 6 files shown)


app/controllers/WhatDoYouWantToDoController.scala
```diff
@@ -120,12 +118,18 @@ class WhatDoYouWantToDoController @Inject()(
...
   private def allowWhatDoYouWantToDo(implicit request: Request[AnyContent], user: AuthedUser): Future[Result] = {
     val nino = user.nino
-    for {
-      hasTaxCodeChanged <- taxCodeChangeService.hasTaxCodeChanged(nino)
-      showJrsTile       <- jrsService.checkIfJrsClaimsDataExist(nino)
-      model             <- whatToDoView(nino, hasTaxCodeChanged, showJrsTile)
-      _                 <- auditNumberOfTaxCodesReturned(nino, showJrsTile)
-    } yield Ok(whatDoYouWantToDoTileView(WhatDoYouWantToDoForm.createForm, model, applicationConfig))
+
+    taxCodeChangeService.hasTaxCodeChanged(nino) flatMap {
```
brendanstraughan, line 122: This just handles taxCodeChangeService.hasTaxCodeChanged returning an error, what if the others in the for loop return an error?

app/uk/gov/hmrc/tai/service/TaxCodeChangeService.scala
```diff
@@ -37,25 +39,25 @@ class TaxCodeChangeService @Inject()(taxCodeChangeConnector: TaxCodeChangeConnec
...
 
-    for {
-      hasTaxCodeChanged <- hasTaxCodeChangedFuture
-      taxCodeMismatch   <- taxCodeMismatchFuture
-    } yield {
-      (hasTaxCodeChanged, taxCodeMismatch) match {
-        case (_: Boolean, TaiSuccessResponseWithPayload(taxCodeMismatch: TaxCodeMismatch)) => {
+    val hasTaxChanged = for {
+      hasTaxCodeChanged <- EitherT(hasTaxCodeChangedFuture)
+      taxCodeMismatch   <- EitherT.right[TaxCodeError](taxCodeMismatchFuture)
```
brendanstraughan, line 48: Never used cats before so will need an intro into it

app/uk/gov/hmrc/tai/service/Error.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package uk.gov.hmrc.tai.service
+
+import uk.gov.hmrc.domain.Nino
+
+sealed trait Error extends Exception
+
+final case class TaxCodeError(nino: Nino, errorMessage: Option[String] = None) extends Error
+
+final case class TaiNotFoundError(message: String) extends Error
```
brendanstraughan, line 25: Do we need to have an error code in here? ie 404

app/controllers/WhatDoYouWantToDoController.scala
```diff
@@ -120,12 +118,18 @@ class WhatDoYouWantToDoController @Inject()(
...
+      case Right(taxCodeChanged) =>
+        for {
+          showJrsTile <- jrsService.checkIfJrsClaimsDataExist(nino)
+          model       <- whatToDoView(nino, taxCodeChanged, showJrsTile)
+          _           <- auditNumberOfTaxCodesReturned(nino, showJrsTile)
+        } yield Ok(whatDoYouWantToDoTileView(WhatDoYouWantToDoForm.createForm, model, applicationConfig))
```
brendanstraughan, line 128: can we do a match on the yield so we can cover all possible error cases?

app/controllers/WhatDoYouWantToDoController.scala
```diff
@@ -18,25 +18,23 @@ package controllers
```
brendanstraughan, line 18: Do these changes need to be done in the backend aswell?

vivekhmrc: Yes but for backend we may need to agree on a convention. Fix the responses returned so that they can be interpreted accordingly by frontend

app/controllers/WhatDoYouWantToDoController.scala
vivekhmrc: Yes they can so we have to handle each exception being thrown around by code

app/uk/gov/hmrc/tai/service/Error.scala
vivekhmrc: Good point. We may

app/controllers/WhatDoYouWantToDoController.scala
vivekhmrc: Not sure I get that but can discuss this later on the call @brendanstraughan

Jamie22Butler (approved):
LGTM 👍

app/uk/gov/hmrc/tai/service/TaxCodeChangeService.scala
sergi-hmrc: if u want i can give an intro to cats one of these days for the whole team

app/controllers/WhatDoYouWantToDoController.scala
```diff
@@ -120,12 +118,18 @@ class WhatDoYouWantToDoController @Inject()(
...
-      showJrsTile       <- jrsService.checkIfJrsClaimsDataExist(nino)
-      model             <- whatToDoView(nino, hasTaxCodeChanged, showJrsTile)
-      _                 <- auditNumberOfTaxCodesReturned(nino, showJrsTile)
-    } yield Ok(whatDoYouWantToDoTileView(WhatDoYouWantToDoForm.createForm, model, applicationConfig))
+
+    taxCodeChangeService.hasTaxCodeChanged(nino) flatMap {
+      case Right(taxCodeChanged) =>
+        for {
+          showJrsTile <- jrsService.checkIfJrsClaimsDataExist(nino)
+          model       <- whatToDoView(nino, taxCodeChanged, showJrsTile)
```
sergi-hmrc, line 126: no big deal but whatToDoView and auditNumberOfTax can be done in parallel: 
```
import cats.implicits._
(model, _)       <- (whatToDoView(nino, taxCodeChanged, showJrsTile), auditNumberOfTaxCodesReturned(nino, showJrsTile)).tupled
```

app/controllers/WhatDoYouWantToDoController.scala
```diff
@@ -120,12 +118,18 @@ class WhatDoYouWantToDoController @Inject()(
...
+          _           <- auditNumberOfTaxCodesReturned(nino, showJrsTile)
+        } yield Ok(whatDoYouWantToDoTileView(WhatDoYouWantToDoForm.createForm, model, applicationConfig))
+      case Left(taxCodeError) =>
+        Future.successful(
+          BadRequest(errorPagesHandler.error4xxPageWithLink(taxCodeError.errorMessage.getOrElse("Tax Code not Found"))))
```
sergi-hmrc, line 131: I don't think the expected behaviour is a BadRequest as the previous code used: 
```final case class TaxCodeError(nino: Nino, errorMessage: Option[String] = None) extends Error``` 
which is not a subset of ```BadRequestException``` 

In order to maintain the same behaviour this should be used: 
```
errorPagesHandler.hodBadRequestResult(nino.toString()) orElse errorPagesHandler.hodInternalErrorResult(
        nino.toString())
        ```

sergi-hmrc: I would propagate the Future[Either[_,_]] in the allowWhatDoYouWantToDo and then use the errorPagesHandler

app/controllers/WhatDoYouWantToDoController.scala
vivekhmrc: We can do that 👍

app/controllers/WhatDoYouWantToDoController.scala
vivekhmrc: Ok 👍 Will incorporate that change. Thanks

PR #861, opened 2022-03-24
Payee 212 cache issues after updating estimated income

Jamie22Butler:
- [ ] Unit tests passing
- [ ] Coverage reached
- [x] Reviewed

### DIFF
--- a/app/controllers/IncomeController.scala
+++ b/app/controllers/IncomeController.scala
@@ -140,35 +140,34 @@ class IncomeController @Inject()(
     })
   }
 
-  def editRegularIncome(): Action[AnyContent] = (authenticate andThen validatePerson).async { implicit request =>
-    implicit val user: AuthedUser = request.taiUser
+  def editRegularIncome(empId: Int): Action[AnyContent] = (authenticate andThen validatePerson).async {
+    implicit request =>
+      implicit val user: AuthedUser = request.taiUser
 
-    journeyCacheService
-      .collectedJourneyValues(
-        Seq(UpdateIncome_PayToDateKey, UpdateIncome_IdKey, UpdateIncome_NameKey),
-        Seq(UpdateIncome_DateKey))
-      .getOrFail flatMap {
-      case (mandatorySeq, optionalSeq) =>
-        val date = Try(optionalSeq.head.map(date => LocalDate.parse(date))) match {
-          case Success(optDate) => optDate
-          case Failure(exception) =>
-            logger.warn(s"Unable to parse updateIncomeDateKey  $exception")
-            None
-        }
+      journeyCacheService
+        .collectedJourneyValues(Seq(UpdateIncome_PayToDateKey, UpdateIncome_NameKey), Seq(UpdateIncome_DateKey))
+        .getOrFail flatMap {
+        case (mandatorySeq, optionalSeq) =>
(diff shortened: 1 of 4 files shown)


app/controllers/IncomeController.scala
```diff
@@ -177,34 +178,67 @@ class IncomeController @Inject()(
...
-            case Some(taxCodeIncome) =>
-              val employmentAmount = EmploymentAmount(taxCodeIncome, employment)
-
-              val vm = ConfirmAmountEnteredViewModel(employment.name, employmentAmount.oldAmount, cachedData(1).toInt)
-              Ok(confirmAmountEntered(vm))
-
-            case _ => throw new RuntimeException(s"Not able to found employment with id $id")
-          }
-        case _ => throw new RuntimeException("Exception while reading employment and tax code details")
+  def confirmRegularIncome(empId: Int): Action[AnyContent] = (authenticate andThen validatePerson).async {
```
sergi-hmrc, line 181: empId moved to the URL so it's needed here as well

app/controllers/IncomeController.scala
```diff
@@ -177,34 +178,67 @@ class IncomeController @Inject()(
...
+    implicit request =>
+      implicit val user: AuthedUser = request.taiUser
+      val nino = user.nino
+
+      (for {
+        cachedData <- journeyCacheService.mandatoryJourneyValueAsInt(UpdateIncome_NewAmountKey).getOrFail.recoverWith {
```
sergi-hmrc, line 187: because the empId comes from the URL it's not cached anymore

test/controllers/IncomeControllerSpec.scala
```diff
@@ -62,6 +62,7 @@ class IncomeControllerSpec extends BaseSpec with JourneyCacheConstants with I18n
   val payToDate = "100"
   val employerId = 1
   val employerName = "Employer Name"
+  val cachePayToDate = Future.successful(Right(100))
 
   val cachedData = Future.successful(Right(Seq(employerId.toString, payToDate)))
```
brendanstraughan, line 67: Should all places that use this now be using cachePayToDate?

app/controllers/IncomeController.scala
```diff
@@ -140,35 +142,34 @@ class IncomeController @Inject()(
...
-        val date = Try(optionalSeq.head.map(date => LocalDate.parse(date))) match {
-          case Success(optDate) => optDate
-          case Failure(exception) =>
-            logger.warn(s"Unable to parse updateIncomeDateKey  $exception")
-            None
-        }
-
-        val employerName = mandatorySeq(2)
-        val payToDate = BigDecimal(mandatorySeq.head)
+  def editRegularIncome(empId: Int): Action[AnyContent] = (authenticate andThen validatePerson).async {
```
brendanstraughan, line 145: I find this very hard to follow. What happens form a high level when the user submits and they then click back? Have UX been asked to see what is best? As we are now using mandatoryJourneyValueAsInt to get from the cache, we can handle a failing retrieval from the cache better using the either, and maybe send them back to the start of the journey which is what is done elsewhere in the service? Do we need to be letting the user submit again?
Here is an example of how I've seen it done before 
https://github.com/hmrc/tai-frontend/blob/afff5a0634235db1fd151f3bba0175afff789c6d/app/controllers/income/UpdateIncomeNextYearController.scala#L256

PR #876, opened 2022-04-26
PAYEE-262: Upgrade the FE Library for Check previous tax years

vivekhmrc:
Signed-off-by: vivekhmrc <vivek.kumar@digital.hmrc.gov.uk>

- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/views/MainTempleteNew.scala.html
+++ b/app/views/MainTempleteNew.scala.html
@@ -136,6 +136,7 @@
 
 @beforeContent = {
 
+
 @if(isGG && isSa) {
     @hmrcAccountMenu(AccountMenu(
         businessTaxAccount = Some(BusinessTaxAccount()),
@@ -158,11 +159,13 @@
     }
 
     @attorneyBanner
-
     @backLinkUrl.map(url =>
-        govukBackLink(BackLink(attributes = Map("id" -> "returnToChooseTaxYearLink"), href = url, content = Text(backLinkContent)))
+        govukBackLink(BackLink(classes = "js-visible-back",
+        attributes = Map("id" -> "returnToChooseTaxYearLink"),
+        href = url,
+        content = Text(backLinkContent)))
     )
-}
+ }
 }
 
 
--- a/app/views/MainTempleteNew.scala.html
+++ b/app/views/MainTempleteNew.scala.html
@@ -161,9 +161,7 @@
     @attorneyBanner
     @backLinkUrl.map(url =>
         govukBackLink(BackLink(classes = "js-visible-back",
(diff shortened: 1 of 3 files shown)


sergi-hmrc:
I've noticed the prototype shows `You can also [check how much Income Tax you paid ](http://localhost:9416/tax-you-paid/status)for 6 April 2021 to 5 April 2022.` for all views but we only show that for the previous year. I'm not sure about which is the expected behaviour:

app/views/TaxCodeDetailsPreviousYearsView.scala.html
```diff
@@ -20,45 +20,54 @@
...
 }
 
-@mainTemplate(
+@main(
     title = taxCodeViewModel.title,
     authedUser = Some(user),
-    pageHeader = Some(header),
-    articleClasses = Some("full-width")){
+    backLinkContent=messages("tai.back-link.upper"),
+    backLinkUrl = Some("javascript:history.go(-1)"),
```
sergi-hmrc, line 41: should we use javascript or `Some(routes.WhatDoYouWantToDoController.whatDoYouWantToDoPage.url),
`?

vivekhmrc: 👍  Incorporated the change

sergi-hmrc (approved):

app/views/TaxCodeDetailsPreviousYearsView.scala.html
brendanstraughan: is it worth checking with ux to see if this is the page we want to be sending the user back to?

app/views/MainTempleteNew.scala.html
```diff
@@ -158,11 +159,11 @@
     }
 
     @attorneyBanner
-
     @backLinkUrl.map(url =>
-        govukBackLink(BackLink(attributes = Map("id" -> "returnToChooseTaxYearLink"), href = url, content = Text(backLinkContent)))
+        govukBackLink(BackLink(classes = "js-visible-back",
```
brendanstraughan, line 163: Not sure if it makes a difference but I done this with Mas and he said to use "js-visible"?

vivekhmrc: I rechecked and in the PROD the same is being used. Will recheck with Mas on return though.

app/views/TaxCodeDetailsPreviousYearsView.scala.html
vivekhmrc: Verified 👍

PR #877, opened 2022-04-26
Payee 202 fe library upgrade coronavirus job retention scheme

Jamie22Butler:
- [x] Unit tests passing
- [ ] Coverage reached
- [x]  HTML Validator passing
- [x] Browser Accessibility Check
- [ ] Reviewed

### DIFF
--- a/app/views/JrsClaimSummaryView.scala.html
+++ b/app/views/JrsClaimSummaryView.scala.html
@@ -34,8 +34,8 @@
 @main(
     title = messages("check.jrs.claims.title"),
     pageTitle = Some(Messages("tai.service.navTitle")),
-    backLinkUrl = None,
-    backLinkContent = Messages("tai.returnToChooseTaxYear"),
+    backLinkUrl = Some(routes.WhatDoYouWantToDoController.whatDoYouWantToDoPage.url),
+    backLinkContent = Messages("check.jrs.back.link"),
 ){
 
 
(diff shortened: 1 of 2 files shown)


app/views/JrsClaimSummaryView.scala.html
```diff
@@ -22,73 +22,79 @@
...
 
 @employerMessageKey = @{viewModel.employerMessageKey}
 
-@mainTemplate(
-    title = messages("check.jrs.claims.title")
+
+@main(
+    title = messages("check.jrs.claims.title"),
+    pageTitle = Some(Messages("tai.service.navTitle")),
+    backLinkUrl = None,
```
brendanstraughan, line 37: What happens when the user clicks to go back if we havent got a URL?

Jamie22Butler: This page is unique in the case that the back link is handled differently, on lines 48 - 55. Whether we want to handle this in a uniform way, that's up for discussion; however, I did not want to make changes which could affect the layout.
The user navigates back to the whatDoYouWantToDoPage.

So there is a back button, it's just not handled by the main template :)

brendanstraughan: how come we cant have that back link in here? e.g. 
backLinkUrl = routes.WhatDoYouWantToDoController.whatDoYouWantToDoPage.url,
backLinkContent = Messages("check.jrs.back.link")

Jamie22Butler: Spoken with Vivek, opted to use the backLink generated from the mainTemplate. 

removed the HTML within the View, as both of them would generate two back links on the page.

brendanstraughan (approved):

PR #898, opened 2022-05-23
Payee 359 employer journey 002

mahowarth:
- [x] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/views/employments/AddEmploymentPayrollNumberFormView.scala.html
+++ b/app/views/employments/AddEmploymentPayrollNumberFormView.scala.html
@@ -19,8 +19,11 @@
 @import uk.gov.hmrc.play.views.html.helpers.FormWithCSRF
 @import uk.gov.hmrc.tai.forms.employments.AddEmploymentPayrollNumberForm
 @import uk.gov.hmrc.tai.viewModels.employments.PayrollNumberViewModel
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+@import uk.gov.hmrc.govukfrontend.views.html.components.implicits._
+@import scala.concurrent.ExecutionContext
 
-@this(mainTemplate: MainTemplate, form: FormWithCSRF)
+@this(main: MainTempleteNew, form: FormWithCSRF, govukRadios: GovukRadios, govukButton: GovukButton, govukInput : GovukInput, govukErrorSummary: GovukErrorSummary)
 
 @(payrollNumberForm: Form[AddEmploymentPayrollNumberForm], payrollNumberViewModel: PayrollNumberViewModel)(implicit request: Request[_], messages: Messages, user: controllers.auth.AuthedUser, templateRenderer: uk.gov.hmrc.renderer.TemplateRenderer, ec: scala.concurrent.ExecutionContext)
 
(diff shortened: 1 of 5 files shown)


Jamie22Butler (approved):
LGTM 👍

app/views/employments/AddEmploymentPayrollNumberFormView.scala.html
```diff
@@ -14,64 +14,87 @@
...
-    pageHeader = Some(header),
-    articleClasses = None
+    backLinkContent = Messages("tai.back-link.upper")
 ) {
+    <div class = "govuk-grid-row">
+        <div class = "govuk-grid-column-two-thirds">
+            @if(payrollNumberForm.errors.nonEmpty) {
+                @govukErrorSummary(ErrorSummary(errorList = payrollNumberForm.errors.asTextErrorLinks, title = Text(messages("tai.income.error.form.summary"))))
+            }
+            @header
```
masahmad, line 49: do we need header ? - As the H1 heading is displayed in the legend on these journeys with forms ?

test/views/html/employments/AddEmploymentPayrollNumberFormViewSpec.scala
```diff
@@ -62,9 +62,9 @@ class AddEmploymentPayrollNumberFormViewSpec extends TaiViewSpec with FormValues
           "tai.addEmployment.employmentPayrollNumber.error.selectOption")
         val formWithErrors: Form[AddEmploymentPayrollNumberForm] = AddEmploymentPayrollNumberForm.form
           .withError(AddEmploymentPayrollNumberForm.PayrollNumberChoice, noPayrollNumberChooseError)
-        def view: Html = add_employment_payroll_number_form(formWithErrors, payrollNumberViewModel)
+        val view = add_employment_payroll_number_form(formWithErrors, payrollNumberViewModel)
```
brendanstraughan, line 65: just out of curiosity why was this changed to a val?

app/views/employments/AddEmploymentPayrollNumberFormView.scala.html
```diff
@@ -14,64 +14,87 @@
...
+                ).withFormField(payrollNumberForm(AddEmploymentPayrollNumberForm.PayrollNumberChoice)))
 
-        <div id="payrollNumberEntry-container" class="@if(payrollNumberForm(AddEmploymentPayrollNumberForm.PayrollNumberChoice).value.getOrElse("") != AddEmploymentPayrollNumberForm.YesValue){js-hidden} data-payrollNumberEntry">
-            @input(
-                field = payrollNumberForm(AddEmploymentPayrollNumberForm.PayrollNumberEntry),
-                labelText = Messages("tai.addEmployment.employmentPayrollNumber.label"),
-                classes = Seq("form-control input--cleared"),
-                panel = true
-            )
+                @govukButton(Button(content = Text(messages("tai.continue"))))
```
brendanstraughan, line 92: does this need to say submit like it did before?

sergi-hmrc: so in the messages file: tai.continue == tai.submit  == "Continue"
imo being that both of them point to continue I'd use continue but again buttons are submitted... I'm happy either way

test/views/html/employments/AddEmploymentPayrollNumberFormViewSpec.scala
sergi-hmrc: because it's only called once and it's not reused outside the scope of this test

app/views/employments/AddEmploymentPayrollNumberFormView.scala.html
sergi-hmrc: the only problem I see is that this view has the header on top of the list, should the list go as part of the Fieldset?

PR #901, opened 2022-05-25
PAYEE-360: updates CanWeContactByPhoneViewModel to contain pageTitle.…

sergi-hmrc:
… Updated CanWeContactByPhoneView.scala.html to use new GDS version

- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/controllers/employments/AddEmploymentController.scala
```diff
@@ -79,7 +79,8 @@ class AddEmploymentController @Inject()(
       messages("tai.canWeContactByPhone.title"),
       controllers.employments.routes.AddEmploymentController.addEmploymentPayrollNumber().url,
       controllers.employments.routes.AddEmploymentController.submitTelephoneNumber().url,
-      controllers.employments.routes.AddEmploymentController.cancel().url
+      controllers.employments.routes.AddEmploymentController.cancel().url,
+      Some(messages("tai.service.navTitle")),
```
brendanstraughan, line 83: If we are just passing in a Message key can we not just put this straight into the view to save us adding a new constructor parameter to the viewModel

sergi-hmrc: I did it this way because I saw it was a resource that was reused across different controllers but I guess they're all for current year (add employment, update employment, add pension and remove company benefits)

Jamie22Butler (approved):
LGTM 👍

app/controllers/employments/AddEmploymentController.scala
sergi-hmrc: as long as they're all `Check your Income Tax` it should be fine, if it's used by any other controller in the future that is not part of `Check your Income Tax` then we will need to pass the pageTitle

brendanstraughan (approved):

PR #905, opened 2022-05-27
PAYEE-365 add govukDateInput component , gds styles,  fix tests

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/views/employments/EndEmploymentView.scala.html
+++ b/app/views/employments/EndEmploymentView.scala.html
@@ -15,41 +15,88 @@
  *@
 
 @import includes.forms._
-@import includes.{cancelLink, ptaHeader}
+@import includes.{cancelLink}
 @import java.time.LocalDate
-@import uk.gov.hmrc.play.views.html.helpers._
+
 @import uk.gov.hmrc.tai.viewModels.employments.EmploymentViewModel
 
 @import scala.concurrent.ExecutionContext
+@import uk.gov.hmrc.govukfrontend.views.html.components.GovukDateInput
 
-@this(mainTemplate: MainTemplate, form: FormWithCSRF)
+@import uk.gov.hmrc.govukfrontend.views.html.components._
+@import uk.gov.hmrc.govukfrontend.views.html.components.implicits._
+@import uk.gov.hmrc.hmrcfrontend.views.config.HmrcPageHeadingLegend
+@import uk.gov.hmrc.hmrcfrontend.views.config.HmrcSectionCaption
 
-@(dateForm: Form[LocalDate], viewmodel: EmploymentViewModel)(implicit request: Request[_], messages: Messages, user: controllers.auth.AuthedUser, templateRenderer: uk.gov.hmrc.renderer.TemplateRenderer,  ec: ExecutionContext)
+@this(mainTemplate: MainTempleteNew, form: FormWithCSRF,  govukDateInput : GovukDateInput, govukButton: GovukButton, govukErrorSummary: GovukErrorSummary)
 
-@header = @{
-    ptaHeader(
-        formForErrorSummary = Some(dateForm),
-        displayBackLink = true,
-        mainHeadingText = Messages("tai.endEmployment.endDateForm.title", viewmodel.employerName),
-        preHeadingText = Messages("tai.endEmployment.preHeadingText")
-    )
+@(dateForm: Form[LocalDate], viewmodel: EmploymentViewModel)(implicit request: Request[_], messages: Messages, user: controllers.auth.AuthedUser,  ec: ExecutionContext)
+
+
+@hint = {
+    <div>
+        <span id="date-you-left-hint" class="govuk-!-font-weight-bold govuk-!-display-block" >@Messages("tai.endEmployment.endDateForm.label" , viewmodel.employerName)</span>
+        <span id="date-example-hint">@Messages("tai.label.date.example")</span>
+    </div>
 }
 
-@mainTemplate(title = Messages("tai.endEmployment.endDateForm.pagetitle"),
-      authedUser = Some(user),
-      pageHeader = Some(header),
-      articleClasses = None) {
+@mainTemplate(
+    title = Messages("tai.endEmployment.endDateForm.pagetitle"),
+    pageTitle = Some(Messages("tai.service.navTitle")),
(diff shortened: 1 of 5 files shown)


app/views/employments/EndEmploymentView.scala.html
```diff
@@ -15,41 +15,88 @@
...
+
+@this(mainTemplate: MainTempleteNew, form: FormWithCSRF,  govukDateInput : GovukDateInput, govukButton: GovukButton, govukErrorSummary: GovukErrorSummary)
 
-@this(mainTemplate: MainTemplate, form: FormWithCSRF)
+@(dateForm: Form[LocalDate], viewmodel: EmploymentViewModel)(implicit request: Request[_], messages: Messages, user: controllers.auth.AuthedUser,  ec: ExecutionContext)
 
-@(dateForm: Form[LocalDate], viewmodel: EmploymentViewModel)(implicit request: Request[_], messages: Messages, user: controllers.auth.AuthedUser, templateRenderer: uk.gov.hmrc.renderer.TemplateRenderer,  ec: ExecutionContext)
 
-@header = @{
-    ptaHeader(
```
brendanstraughan, line 30: was the header meant to be removed?

app/views/employments/EndEmploymentView.scala.html
```diff
@@ -15,41 +15,88 @@
...
 
-@mainTemplate(title = Messages("tai.endEmployment.endDateForm.pagetitle"),
-      authedUser = Some(user),
-      pageHeader = Some(header),
-      articleClasses = None) {
+@mainTemplate(
+    title = Messages("tai.endEmployment.endDateForm.pagetitle"),
+    pageTitle = Some(Messages("tai.service.navTitle")),
+    authedUser = Some(user),
+    backLinkId = "backLink",
```
brendanstraughan, line 47: I think this will fail the build as backLinkContent is now a required field

test/uk/gov/hmrc/tai/util/viewHelpers/TaiViewSpec.scala
```diff
@@ -74,7 +74,7 @@ trait TaiViewSpec extends BaseSpec with JsoupMatchers {
       }
     }
     "have an h1 header consisting of the main heading text" in {
-      doc must haveHeadingWithText(mainHeaderText)
+      // doc must haveHeadingWithText(mainHeaderText)
```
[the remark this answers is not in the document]
masahmad: Ooops - checking now

app/views/employments/EndEmploymentView.scala.html
masahmad: yes its not used in pages where we are using a form as the H1 heading is handled by the Form field legend - this is to prevent duplication of  the heading and was originally an issue reported from the accessibility team. Also as we're unlikely to have the print button on the form pages  - there is no need for the header anymore

app/views/employments/EndEmploymentView.scala.html
```diff
@@ -45,6 +45,7 @@
     pageTitle = Some(Messages("tai.service.navTitle")),
     authedUser = Some(user),
     backLinkId = "backLink",
+    backLinkContent = Messages("tai.back-link.upper")
```
brendanstraughan, line 48: needs to be a Some() I believe

test/uk/gov/hmrc/tai/util/viewHelpers/TaiViewSpec.scala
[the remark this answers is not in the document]
masahmad: I've uncommented the test  and pushed however (it fails)  we have to wait until ticket PAYEE-359 is merged as that has an update to the same test that will solve the issue.

https://github.com/hmrc/tai-frontend/pull/898/files#diff-b6b05415cca2409ac354167559257c7557055052ec49b74dac83b6d77c993340L63

app/views/employments/EndEmploymentView.scala.html
masahmad: thanks added

app/views/employments/EndEmploymentView.scala.html
masahmad: added

app/views/employments/EndEmploymentView.scala.html
```diff
@@ -15,41 +15,89 @@
...
 
-@header = @{
-    ptaHeader(
-        formForErrorSummary = Some(dateForm),
-        displayBackLink = true,
-        mainHeadingText = Messages("tai.endEmployment.endDateForm.title", viewmodel.employerName),
-        preHeadingText = Messages("tai.endEmployment.preHeadingText")
-    )
+@hint = {
+    <div>
```
sergi-hmrc, line 37: is this div needed? i think the Hint puts everything inside one (i might be wrong)

app/views/employments/EndEmploymentView.scala.html
```diff
@@ -15,41 +15,89 @@
...
+
+        @govukDateInput(DateInput(
+            id = "tellUsAboutEmploymentForm",
+            namePrefix = Some("tellUsAboutEmploymentForm"),
+            hint = Some(Hint(content = HtmlContent(hint))),
+            items = Seq(
+                InputItem(
+                    name= "day",
+                    label = Some(messages("tai.label.day")),
+                    classes = s"govuk-input--width-2 ${if(dateForm.hasErrors)  " govuk-input--error" else "" }",
```
sergi-hmrc, line 70: are classes needed? i think if you use errorMessage with a filter by key should highlight only the one that has an error

masahmad: discussed with Sergi - we can ignore this as it may not be possible with this component.

sergi-hmrc: yup here's an example: https://github.com/hmrc/manage-transit-movements-amendment-frontend/blob/19e8c66587ba6647cd763ecc9a61a90a6922083c/app/viewmodels/govuk/DateFluency.scala

InputItem doesn't allow an errorMessage so LGTM

sergi-hmrc (approved):

brendanstraughan (approved):

PR #922, opened 2022-06-21
PAYEE-377 Update taxable income 001

mahowarth:
- [x] Unit tests passing
- [ ] Coverage reached
- [x] Reviewed

### DIFF
--- a/app/views/incomes/EstimatedPayLandingPageView.scala.html
+++ b/app/views/incomes/EstimatedPayLandingPageView.scala.html
@@ -49,8 +49,8 @@ ec: scala.concurrent.ExecutionContext)
         </div>
         <div class="govuk-grid-column-two-thirds">
             <p class="govuk-body">@messages("tai.incomes.landing.intro")</p>
-            <a id="updateEstimatedPayTotalIncome" aria-hidden="true" class="govuk-visually-hidden">@messages("tai.incomes.landing.employment.totalIncome", totalEstimatedIncome)</a>
-            <p id="updateEstimatedPayLink" aria-hidden="true" class="govuk-visually-hidden"><a href="@routes.EstimatedIncomeTaxController.estimatedIncomeTax">@messages("tai.incomes.landing.employment.affect")</a></p>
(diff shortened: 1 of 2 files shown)


app/views/incomes/EstimatedPayLandingPageView.scala.html
```diff
@@ -30,47 +30,54 @@
...
-        <p id="updateEstimatedPayLink" aria-hidden="true"><a href="@routes.EstimatedIncomeTaxController.estimatedIncomeTax">@messages("tai.incomes.landing.employment.affect")</a></p>
-        <p>
-            @if(isPension) {
-                @messages("tai.incomes.landing.pension.ended", incomeName)
-                @link(
-                        url=appConfig.incomeFromEmploymentPensionLinkUrl,
-                        copy=messages("tai.incomes.landing.pension.ended.link"),
-                        id=Some("updatePension")
+    <div class="govuk-grid-row">
+        <div class = "govuk-grid-column-two-thirds">
```
sergi-hmrc, line 47: nitpick: you could also put header on line 51 right above the first `p`

sergi-hmrc (approved):

app/views/incomes/EstimatedPayLandingPageView.scala.html
```diff
@@ -30,47 +30,54 @@
...
+            @header
+        </div>
+        <div class="govuk-grid-column-two-thirds">
+            <p class="govuk-body">@messages("tai.incomes.landing.intro")</p>
+            <a id="updateEstimatedPayTotalIncome" aria-hidden="true" class="govuk-visually-hidden">@messages("tai.incomes.landing.employment.totalIncome", totalEstimatedIncome)</a>
```
masahmad, line 52: These 2 lines where we have visually hidden classes means they are for screen readers only - however they also have aria-hidden true which hides it from screen readers so there is a contradiction here I think  ???

Jamie22Butler (approved):
LGTM 👍

PR #923, opened 2022-06-22
PAYEE-466 - display unavailable if tax code couldn't be retrieved

sergi-hmrc:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/IncomeTaxHistoryController.scala
+++ b/app/controllers/IncomeTaxHistoryController.scala
@@ -49,8 +49,9 @@ class IncomeTaxHistoryController @Inject()(
 )(implicit ec: ExecutionContext, templateRenderer: TemplateRenderer)
     extends TaiBaseController(mcc) {
 
-  private def getIncomeTaxYear(nino: Nino, taxYear: TaxYear)(implicit hc: HeaderCarrier): Future[IncomeTaxYear] = {
-    val futureMaybeTaxCodes = taxAccountService.taxCodeIncomesV2(nino, taxYear).map(_.toOption)
+  def getIncomeTaxYear(nino: Nino, taxYear: TaxYear)(implicit hc: HeaderCarrier): Future[IncomeTaxYear] = {
+    val futureMaybeTaxCodes =
+      taxAccountService.taxCodeIncomesV2(nino, taxYear).map(_.toOption).recover { case _ => None }
     val futureEmployments = employmentService.employments(nino, taxYear)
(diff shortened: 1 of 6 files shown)


conf/messages.cy
```diff
@@ -567,6 +567,7 @@ tai.incomeTax.history.endDate = Dyddiad dod i ben
 tai.incomeTax.history.taxableIncome = Incwm trethadwy
 tai.incomeTax.history.incomeTaxPaid = Treth incwm a dalwyd
 tai.incomeTax.history.taxCode = Cod treth
+tai.incomeTax.history.unavailable = xyz
```
masahmad, line 570: is that the welsh translation "xyx"- how do you pronounce it :) ?

vivekhmrc: :-0)

masahmad (approved):
LGTM - pending the welsh update

app/controllers/IncomeTaxHistoryController.scala
```diff
@@ -49,8 +49,9 @@ class IncomeTaxHistoryController @Inject()(
 )(implicit ec: ExecutionContext, templateRenderer: TemplateRenderer)
     extends TaiBaseController(mcc) {
 
-  private def getIncomeTaxYear(nino: Nino, taxYear: TaxYear)(implicit hc: HeaderCarrier): Future[IncomeTaxYear] = {
-    val futureMaybeTaxCodes = taxAccountService.taxCodeIncomesV2(nino, taxYear).map(_.toOption)
+  def getIncomeTaxYear(nino: Nino, taxYear: TaxYear)(implicit hc: HeaderCarrier): Future[IncomeTaxYear] = {
+    val futureMaybeTaxCodes =
+      taxAccountService.taxCodeIncomesV2(nino, taxYear).map(_.toOption).recover { case _ => None }
```
brendanstraughan, line 54: does .recover just recover exceptions? If so what happens when taxCodeIncomesV2 returns successful but no tax code inside. I say this because for my ticket I removed a load of data from the stubs and it only threw an exception for employmentService.employments. I think taxAccountService.taxCodeIncomesV2 is handled gracefully so not sure if the .recover will be invoked?

brendanstraughan (approved):

conf/messages.cy
sergi-hmrc: updated!

PR #935, opened 2022-07-05
PAYEE-387 upgrade fe  estimated taxable income journey - 011

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/views/incomes/BonusPaymentAmountView.scala.html
+++ b/app/views/incomes/BonusPaymentAmountView.scala.html
@@ -16,8 +16,8 @@
 
 @import includes.ptaHeader
 @import includes.cancelLink
-@import includes.forms.{input, submit}
-@*@import uk.gov.hmrc.play.views.html.helpers._*@
+
+
 @import uk.gov.hmrc.tai.util.TaxYearRangeUtil
 @import uk.gov.hmrc.tai.model.domain.income.IncomeSource
 @import scala.concurrent.ExecutionContext
@@ -25,7 +25,6 @@
 
 @import uk.gov.hmrc.govukfrontend.views.html.components._
 @import uk.gov.hmrc.govukfrontend.views.html.components.implicits._
-@import uk.gov.hmrc.hmrcfrontend.views.config.HmrcPageHeadingLegend
 @import uk.gov.hmrc.hmrcfrontend.views.config.HmrcSectionCaption
 @import uk.gov.hmrc.hmrcfrontend.views.config.HmrcPageHeadingLabel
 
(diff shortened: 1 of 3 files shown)


app/views/incomes/BonusPaymentAmountView.scala.html
```diff
@@ -33,33 +39,57 @@
...
+
+
 @mainTemplate(
     title = messages("tai.bonusPaymentsAmount.title",TaxYearRangeUtil.currentTaxYearRangeBetweenDelimited),
-    articleClasses = None,
     authedUser = Some(user),
-    pageHeader = Some(header),
-    employerName = Some(employer.name)
+    backLinkContent = Some(Messages("tai.back-link.upper")),
+    backLinkId = "backLink"
```
sergi-hmrc, line 56: why backLink?

app/views/incomes/BonusPaymentAmountView.scala.html
```diff
@@ -33,33 +39,57 @@
         formForErrorSummary = Some(bonusOvertimeAmountForm),
         displayBackLink = true,
```
brendanstraughan, line 40: should this be false?

app/views/incomes/BonusPaymentAmountView.scala.html
masahmad: why not? tests are probably expecting id of "backLink"  :)

app/views/incomes/BonusPaymentAmountView.scala.html
masahmad: yes - changed it -  btw  old backLink has no effect if showGdsVersion is set to true in header

app/views/incomes/BonusPaymentAmountView.scala.html
```diff
@@ -33,33 +39,57 @@
...
 
+            @govukButton(Button(content = Text(messages("tai.selectIncome.link.text"))))
+
 
-            @submit()
             @cancelLink(Call("GET",controllers.routes.IncomeController.cancel(employer.id).url))
         }
     </div>
+
+        </div></div>
```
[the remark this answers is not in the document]
masahmad: yes  done!

app/views/incomes/BonusPaymentAmountView.scala.html
sergi-hmrc: should we change the defaulted one to be backLink then?

masahmad: i was thinking the same - there appears more "backlink" id then the current default one  - maybe another ticket for the backlog

sergi-hmrc (approved):

PR #963, opened 2022-07-26
PAYEE-401: update what do you want to tell us view

sergi-hmrc:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/views/pensions/update/WhatDoYouWantToTellUsView.scala.html
+++ b/app/views/pensions/update/WhatDoYouWantToTellUsView.scala.html
@@ -14,44 +14,61 @@
  * limitations under the License.
  *@
 
-@import includes.forms.{submit, textArea}
-@import includes.{cancelLink, ptaHeader}
+@import includes.forms.submit
+@import includes.cancelLink
 @import uk.gov.hmrc.tai.viewModels.employments.EmploymentViewModel
 @import uk.gov.hmrc.play.views.html.helpers.FormWithCSRF
 @import uk.gov.hmrc.tai.forms.employments.UpdateEmploymentDetailsForm
 @import uk.gov.hmrc.tai.forms.pensions.WhatDoYouWantToTellUsForm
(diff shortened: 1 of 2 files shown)


masahmad (approved):

app/views/pensions/update/WhatDoYouWantToTellUsView.scala.html
```diff
@@ -14,44 +14,61 @@
...
-    )
-}
 
 @mainTemplate(
     title = Messages("tai.updatePension.whatDoYouWantToTellUs.pagetitle"),
+    pageTitle = Some(messages("tai.service.navTitle")),
     authedUser = Some(user),
-    pageHeader = Some(header),
-    articleClasses = None) {
+    backLinkContent = Some(Messages("tai.back-link.upper"))
```
brendanstraughan, line 43: if we dont pass in backLinkId it will default to "returnToChooseTaxYearLink", do we want that for this page?

sergi-hmrc: my question is why does the default template have that instead of backLink? :D but yes maybe i should change that to backLink (but that means do it in all places u.u)

sergi-hmrc: also there is a problem calling it backlink because then there's some JS code that overrides any backLinkId into the JS function

brendanstraughan: hmm shall we say lets make another ticket to change the mainTemplate default ?

brendanstraughan: Going to merge an resolve as part of https://jira.tools.tax.service.gov.uk/browse/PAYEE-533

brendanstraughan (approved):

PR #967, opened 2022-07-29
PAYEE-520 update banner fix

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


vivekhmrc (approved):
LGTM 👍

app/views/IncomeSourceSummaryView.scala.html
```diff
@@ -116,9 +116,21 @@
         }
     }
 
+
     @if(model.estimatedPayJourneyCompleted){
-        <div class="tai-progress-panel" id="estimatedPayCompletionBanner">
-            <h2 class="heading-medium flush-bottom flush--top" id="incomeSummary-banner-heading">@messages("tai.estimatedIncome.confirmation.banner.heading")</h2>
```
brendanstraughan, line 121: are the ID's here needed as they have been removed? Im wondering if they may be used in the AT's?

masahmad: that's a good point - the AT's will probably need to be amended regardless as the new GDS banner does not use H2 and the parent div is different too.  I would wait andsee

brendanstraughan (approved):

PR #998, opened 2022-08-31
PAYEE-562: Capitalised letter

PerrettJ4:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


conf/messages
```diff
@@ -566,7 +566,7 @@ tai.incomeTax.history.employerReference = Employer PAYE reference
 tai.incomeTax.history.startDate = Start date
 tai.incomeTax.history.endDate = End date
 tai.incomeTax.history.taxableIncome = Taxable income
-tai.incomeTax.history.incomeTaxPaid = Income tax paid
+tai.incomeTax.history.incomeTaxPaid = Income Tax paid
```
brendanstraughan, line 569: do we need to make this change in the Welsh as well?

PerrettJ4: good point, I believe we do, I will amend today

PR #1005, opened 2022-09-01
FE updates and remove unused main template param

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/views/MainTempleteNew.scala.html
+++ b/app/views/MainTempleteNew.scala.html
@@ -85,7 +85,7 @@
         headerContainerClasses: String = Header.defaultObject.containerClasses,
         pagePrintable: Boolean = false,
         pagePrintName: Option[String] = None,
-        hidePtaAccountNav: Boolean = false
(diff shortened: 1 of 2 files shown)


vivekhmrc (approved):

app/views/MainTempleteNew.scala.html
```diff
@@ -137,21 +137,23 @@
 
 @beforeContent = {
 
-
```
sergi-hmrc, line 140: just as personal preference, if the whole purpose of a boolean is to do !boolean, i'd rather change the name so the ! isn't necessary:
if(showPtaAccountNav)

sergi-hmrc: also this would align with showGds and other stuff that is already present

masahmad: agree 100% i changed it (showPtaAccountNav  :)

sergi-hmrc (approved):

PR #1014, opened 2022-09-06
 PAYEE-598: update IncorrectTaxableIncomeView.scala.html

sergi-hmrc:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


masahmad (approved):
looks good  - approving for now
only 1 point for info only - we now have "backLinkId" set in the new  main template as default so prob  no need to specify it here unless it needs to be different.

app/views/incomes/IncorrectTaxableIncomeView.scala.html
```diff
@@ -22,43 +22,51 @@
...
 @( payYTD: BigDecimal = 0, payDate: java.time.LocalDate, id: Int)(implicit request: Request[_], messages: Messages,
 user: controllers.auth.AuthedUser, templateRenderer: uk.gov.hmrc.renderer.TemplateRenderer, ec: scala.concurrent.ExecutionContext)
 
 @mainTemplate(
     title = messages("tai.estimatedPay.error.incorrectTaxableIncome.title"),
-    articleClasses = Some("full-width text"),
-    authedUser = Some(user)
+    pageTitle = Some(Messages("tai.service.navTitle")),
+    backLinkId = "backLink",
+    backLinkUrl = Some(controllers.income.estimatedPay.update.routes.IncomeUpdateCalculatorController.checkYourAnswersPage.url),
```
brendanstraughan, line 34: There was no back link previously, not sure if we need one?

brendanstraughan: confirmed the new back button is needed

brendanstraughan (approved):

PR #1042, opened 2022-09-29
PAYEE-296 add print feature , update viewModel and tests with username

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/controllers/TaxCodeChangeController.scala
```diff
@@ -62,9 +62,15 @@ class TaxCodeChangeController @Inject()(
       val taxCodeChangeReasons = taxCodeChangeReasonsService
         .combineTaxCodeChangeReasons(iabdTaxCodeChangeReasons, yourTaxFreeAmountComparison.iabdPairs, taxCodeChange)
       val isAGenericReason = taxCodeChangeReasonsService.isAGenericReason(taxCodeChangeReasons)
+      val maybeUserName = Some(request.fullName)
```
brendanstraughan, line 65: very picky, fel free to ignore, but this isnt a "maybe", requst.fullName is a string as we will always have it, I think it only makes sense to have it called "maybe" in the view model where it can be a none. Make sense?

brendanstraughan (approved):

robynHM (approved):

PR #1044, opened 2022-10-03
PAYEE-477: Removed the Runtime Exceptions from the Controllers wherever not needed

vivekhmrc:
- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/IncomeTaxComparisonController.scala
+++ b/app/controllers/IncomeTaxComparisonController.scala
@@ -18,11 +18,8 @@ package controllers
 
 import controllers.actions.ValidatePerson
 import controllers.auth.AuthAction
-
-import javax.inject.Inject
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import uk.gov.hmrc.play.audit.http.connector.AuditConnector
-
 import uk.gov.hmrc.renderer.TemplateRenderer
 import uk.gov.hmrc.tai.config.ApplicationConfig
 import uk.gov.hmrc.tai.connectors.responses.TaiSuccessResponseWithPayload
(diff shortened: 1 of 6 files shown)


app/uk/gov/hmrc/tai/connectors/JourneyCacheConnector.scala
```diff
@@ -36,6 +36,7 @@ class JourneyCacheConnector @Inject()(httpHandler: HttpHandler, servicesConfig:
   def currentCache(journeyName: String)(implicit hc: HeaderCarrier): Future[Map[String, String]] =
     httpHandler.getFromApiV2(cacheUrl(journeyName)).map(_.as[Map[String, String]]) recover {
       case e: HttpException if e.responseCode == NO_CONTENT => Map.empty[String, String]
+      case e: HttpException if e.responseCode == NO_CONTENT => Map.empty[String, String]
```
sergi-hmrc, line 39: this is duplicated, from above isn't it?

app/controllers/IncomeTaxComparisonController.scala
```diff
@@ -118,7 +116,8 @@ class IncomeTaxComparisonController @Inject()(
 
           Ok(mainView(model, applicationConfig))
         }
-        case _ => throw new RuntimeException("Not able to fetch income tax comparision details")
+        case _ =>
```
sergi-hmrc, line 119: Should we return a 500 if any of the services returns a TaiFailureResponse? 

Should we gather more information about why this couldn't happen? Which were the culprits to error etc? 
Have we considered alternatives for the API rather than returning Future[TaiResponse], return Future[Either[Error, Payload]] or Future[Payload] etc? 
Also I think

app/uk/gov/hmrc/tai/connectors/JourneyCacheConnector.scala
vivekhmrc: Yes. I was going to change it  but then left it as decided to restrain to just Controllers. Will remove .

app/controllers/benefits/CompanyBenefitController.scala
```diff
@@ -99,7 +97,8 @@ class CompanyBenefitController @Inject()(
             Ok(updateOrRemoveCompanyBenefitDecision(viewModel))
           }
 
-        case None => throw new RuntimeException("No employment found")
+        case None =>
+          Future.successful(errorPagesHandler.internalServerError(s"No employment found for ${user.nino}"))
```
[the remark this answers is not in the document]
vivekhmrc: Yes.. Or a custom page for these kind of issues that passes an appropriate message to the user

PR #1068, opened 2022-10-26
all return types are displayed for public members

robynHM:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -20,3 +20,5 @@ addSbtPlugin("org.irundaia.sbt" % "sbt-sassify" % "1.5.1")
 addSbtPlugin("net.ground5hark.sbt" % "sbt-concat" % "0.2.0")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-uglify" % "2.0.0")
+
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.10.4")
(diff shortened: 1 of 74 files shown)


project/plugins.sbt
```diff
@@ -20,3 +20,5 @@ addSbtPlugin("org.irundaia.sbt" % "sbt-sassify" % "1.5.1")
 addSbtPlugin("net.ground5hark.sbt" % "sbt-concat" % "0.2.0")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-uglify" % "2.0.0")
+
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.10.4")
```
brendanstraughan, line 24: do we need to get approval for plugins? If so I guess we can always remove this plugin now that we have added all the type annotations

brendanstraughan: cc @vivekhmrc

vivekhmrc: @brendanstraughan / @robynHM I checked and it is being used in a lot other repositories so I think we are good :-)

brendanstraughan (approved):

sergi-hmrc (approved):
excellent job @robynHM! Thanks for adding scalafix :)

PR #1072, opened 2022-10-27
PAYEE-594: Facilitator to drop the cache for acceptance tests

vivekhmrc:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/testOnly/TaiUpdateIncomeController.scala
+++ b/app/controllers/testOnly/TaiUpdateIncomeController.scala
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2022 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.testOnly
+
+import com.google.inject.name.Named
+import controllers.TaiBaseController
+import controllers.actions.ValidatePerson
+import controllers.auth.AuthAction
+import play.api.Logging
+import play.api.mvc._
+import uk.gov.hmrc.renderer.TemplateRenderer
+import uk.gov.hmrc.tai.service.journeyCache.JourneyCacheService
+import uk.gov.hmrc.tai.util.constants._
+
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.ExecutionContext
+
+@Singleton
+class TaiUpdateIncomeController @Inject()(
+  @Named("Update Income") journeyCacheService: JourneyCacheService,
+  authenticate: AuthAction,
+  validatePerson: ValidatePerson,
+  mcc: MessagesControllerComponents,
+  implicit val templateRenderer: TemplateRenderer,
+)(implicit ec: ExecutionContext)
+    extends TaiBaseController(mcc) with JourneyCacheConstants with FormValuesConstants with Logging {
+
+  def delete(empId: Int): Action[AnyContent] = (authenticate andThen validatePerson).async { implicit request =>
+    journeyCacheService.delete() map { _ =>
+      Redirect(controllers.routes.IncomeSourceSummaryController.onPageLoad(empId))
+    }
+  }
+}
--- a/app/controllers/testOnly/TaiUpdateIncomeController.scala
+++ b/app/controllers/testOnly/TaiUpdateIncomeController.scala
@@ -20,11 +20,9 @@ import com.google.inject.name.Named
 import controllers.TaiBaseController
 import controllers.actions.ValidatePerson
 import controllers.auth.AuthAction
-import play.api.Logging
 import play.api.mvc._
 import uk.gov.hmrc.renderer.TemplateRenderer
 import uk.gov.hmrc.tai.service.journeyCache.JourneyCacheService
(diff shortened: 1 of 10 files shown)


app/uk/gov/hmrc/tai/connectors/JourneyCacheConnector.scala
```diff
@@ -84,4 +84,8 @@ class JourneyCacheConnector @Inject()(httpHandler: HttpHandler, servicesConfig:
   def flushWithEmpId(journeyName: String, empId: Int)(implicit hc: HeaderCarrier): Future[TaiResponse] =
     httpHandler.deleteFromApi(cacheUrl(s"$journeyName/$empId")).map(_ => TaiSuccessResponse)
 
+  def testOnlyCacheUrl(journeyName: String) = s"$serviceUrl/tai/test-only/journey-cache/$journeyName"
```
brendanstraughan, line 87: why do we need to make a new route, why cant we just use the existing method for flushing the cache?

app/uk/gov/hmrc/tai/service/journeyCache/JourneyCacheService.scala
```diff
@@ -133,4 +134,7 @@ class JourneyCacheService @Inject()(val journeyName: String, journeyCacheConnect
 
   def flushWithEmpId(empId: Int)(implicit hc: HeaderCarrier): Future[TaiResponse] =
```
brendanstraughan, line 135: can we not just use this?

conf/testOnlyDoNotUseInAppConf.routes
```diff
@@ -0,0 +1,16 @@
...
+# To do so add the following line to the micro-service profile: "-Dapplication.router=testOnlyDoNotUseInAppConf.Routes"
+# To start the micro-service locally using the test routes run the following command: "sbt run -Dapplication.router=testOnlyDoNotUseInAppConf.Routes" 
+
+# Any test-only end-point should be defined here.
+# !!!WARNING!!! Every route defined in this file MUST be prefixed with "/test-only/". This is because NGINX is blocking every uri containing the string "test-only" in production.
+# Failing to follow this rule may result in test routes deployed in production.
+
+# Add all the application routes to the prod.routes file
+# microservice specific routes
+
```
brendanstraughan, line 14: why do we need this file, are there any routes being used in here?

conf/tai.routes
```diff
@@ -165,6 +165,7 @@ POST        /update-income/bonus-overtime-amount/:empId
 ## Regular Income URIs
 GET         /update-income/cancel/:id                                                  controllers.IncomeController.cancel(id: Int)
 GET         /update-income/select-taxable-pay                                          controllers.IncomeController.viewIncomeForEdit
+GET         /test-only/update-income/delete/:id                                        controllers.IncomeController.delete(id: Int)
```
sergi-hmrc, line 168: why not just using the DELETE verb?

sergi-hmrc: Also this should be part of conf/testOnlyDoNotUseInAppConf.routes

sergi-hmrc:
A few things: 
Rather than creating `conf/testOnlyDoNotUseInAppConf.routes` we should reuse `dev.routes`. 
Anything inside `dev.routes` will need a different package notice the `controllers.testOnly`: 
`# Routes for development environments
DELETE         /update-income/:id   controllers.testOnly.IncomeController.delete(id: Int)
`
As described [here](https://stackoverflow.com/questions/20211734/play-framework-split-routes-in-multiple-files-without-sub-projects) it won't work otherwise.

Finally, if the only intention is to delete the cache for testing purposes I don't think we should redirect the user to any page but just do a DELETE against the endpoint and return an OK/No Content etc

app/uk/gov/hmrc/tai/service/journeyCache/JourneyCacheService.scala
vivekhmrc: This is just to keep the test ones separate

conf/testOnlyDoNotUseInAppConf.routes
vivekhmrc: Yes moved the test only route here

conf/tai.routes
vivekhmrc: Get is to facilitate testers. The selenium web driver can easily call Get as its doing at other places. With Delete the testers will have to write custom code for this.

app/uk/gov/hmrc/tai/connectors/JourneyCacheConnector.scala
vivekhmrc: This is just to keep the test ones separate

app/controllers/testOnly/TaiUpdateIncomeController.scala
```diff
@@ -0,0 +1,45 @@
...
+class TaiUpdateIncomeController @Inject()(
+  @Named("Update Income") journeyCacheService: JourneyCacheService,
+  authenticate: AuthAction,
+  validatePerson: ValidatePerson,
+  mcc: MessagesControllerComponents,
+  implicit val templateRenderer: TemplateRenderer
+)(implicit ec: ExecutionContext)
+    extends TaiBaseController(mcc) {
+
+  def delete(empId: Int): Action[AnyContent] = (authenticate andThen validatePerson).async { implicit request =>
```
brendanstraughan, line 40: I thought we said we were only going to make changes to the backend
@vivekhmrc @sergi-hmrc

sergi-hmrc: Agree ^

vivekhmrc: In the current tai-acceptance-tests we don't have a Web Service Client to directly invoke **tai**.  So we need a test route here to facilitate it. Once the **_tai-acceptance-test_** are enhanced to have a Web service client e.g. Akka HTTP or some other then we can remove this. It's captured here

https://jira.tools.tax.service.gov.uk/browse/PAYEE-872

PR #1086, opened 2022-11-03
PAYEE-829: Parallel calls in ExternalServiceRedirectController

Jake-Raffe:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/ExternalServiceRedirectController.scala
+++ b/app/controllers/ExternalServiceRedirectController.scala
@@ -18,6 +18,7 @@ package controllers
 
 import controllers.actions.ValidatePerson
 import controllers.auth.AuthAction
+import cats.implicits._
 import javax.inject.Inject
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import uk.gov.hmrc.tai.service.{AuditService, SessionService}
@@ -35,16 +36,12 @@ class ExternalServiceRedirectController @Inject()(
 
   def auditInvalidateCacheAndRedirectService(serviceAndIFormName: String): Action[AnyContent] =
     (authenticate andThen validatePerson).async { implicit request =>
-      {
-
-        (for {
-          redirectUri <- auditService.sendAuditEventAndGetRedirectUri(request.taiUser.nino, serviceAndIFormName)
-          _           <- sessionService.invalidateCache()
-        } yield {
-          Redirect(redirectUri)
-        }) recover {
-          case _ => errorPagesHandler.internalServerError("Unable to audit and redirect")
-        }
+      (
+        auditService.sendAuditEventAndGetRedirectUri(request.taiUser.nino, serviceAndIFormName),
+        sessionService.invalidateCache()).mapN {
+        case (redirectUri, _) => Redirect(redirectUri)
+      } recover {
+        case _ => errorPagesHandler.internalServerError("Unable to audit and redirect")
       }
     }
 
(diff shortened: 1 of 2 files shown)


app/controllers/ExternalServiceRedirectController.scala
```diff
@@ -35,16 +36,12 @@ class ExternalServiceRedirectController @Inject()(
...
-        (for {
-          redirectUri <- auditService.sendAuditEventAndGetRedirectUri(request.taiUser.nino, serviceAndIFormName)
-          _           <- sessionService.invalidateCache()
-        } yield {
-          Redirect(redirectUri)
-        }) recover {
-          case _ => errorPagesHandler.internalServerError("Unable to audit and redirect")
-        }
+      (
+        auditService.sendAuditEventAndGetRedirectUri(request.taiUser.nino, serviceAndIFormName),
```
sergi-hmrc, line 40: Should this one be parallelised? If the audit event can't be send previously we would invalidate the cache. Now this behaviour won't happen, because no matter what we will always invalidate the Cache.

sergi-hmrc: Actually, after looking at the `sendAuditEventAndGetRedirectUri` method it can't fail as it always returns: 
`Future.successful(redirectUri)`

I think instead it should do: 
`sendIformRedirectUriAuditEvent(nino, fetchPath(request), auditEventName).map{ _ => redirectUri} `

And then the behaviour I explain would happen... Otherwise they both do the same thing :D

brendanstraughan: @vivekhmrc whats your thoughts on this. Should we change this to use sendIformRedirectUriAuditEvent instead?

vivekhmrc: @sergi-hmrc / @brendanstraughan Morning guys as a thumb rule the audit event is always in the end to capture the completion of the operations (success or failure). 

I think the logic for auditInvalidateCacheAndRedirectService should have been that the cache is dropped first and then the audit event written. Probably that's the reason why `sendAuditEventAndGetRedirectUri` is always returning `Future.successful(redirectUri)` regardless of the outcome of [auditConnector.sendEvent](url) `createAndSendAuditEvent`

sergi-hmrc: Cool. In that case, the code should be sequential: 
```
for {
          _           <- sessionService.invalidateCache()
          redirectUri <- auditService.sendAuditEventAndGetRedirectUri(request.taiUser.nino, serviceAndIFormName)
        } yield Redirect(redirectUri)
```

Finally the Future.successful is more likely to be an error of misunderstanding how futures work than anything else (after seeing some other code in tai...). Should we map the redirect on the result so it's sequential rather than fire and forget?

robynHM (approved):

app/controllers/ExternalServiceRedirectController.scala
```diff
@@ -36,10 +37,9 @@ class ExternalServiceRedirectController @Inject()(
   def auditInvalidateCacheAndRedirectService(serviceAndIFormName: String): Action[AnyContent] =
     (authenticate andThen validatePerson).async { implicit request =>
       {
-
         (for {
-          redirectUri <- auditService.sendAuditEventAndGetRedirectUri(request.taiUser.nino, serviceAndIFormName)
           _           <- sessionService.invalidateCache()
+          redirectUri <- auditService.sendAuditEventAndGetRedirectUri(request.taiUser.nino, serviceAndIFormName)
```
sergi-hmrc, line 42: might be worth modifying `sendAuditEventAndGetRedirectUri` instead of returning Future.successful it uses a `map`

PR #1133, opened 2022-12-05
PAYEE-940 update list content on tile

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


Jake-Raffe (approved):

app/views/includes/forms/card.scala.html
```diff
@@ -37,7 +39,16 @@
         @heading
       }
     </@headingTag>
-    <p> @bodyContent</p>
+    @if(isBodyContentList){
```
brendanstraughan, line 42: not sure about this so @sergi-hmrc correct me if wrong but would doing 
@if(bodyContentList.nonEmpty)
be better as then we wouldnt have to add extra parameters

sergi-hmrc: isn't the intention to always have a list? then why do we keep a list and a body content? 
Shouldn't it just be `bodyContentList: Seq[String] = Seq.empty`

sergi-hmrc: for now, and as @brendanstraughan said, just use nonEmpty. no need to add another parameter

masahmad: cool updated

sergi-hmrc (approved):
lgtm

vivekhmrc:
The content suggested on this story has been superseded by another iteration of the prototype so this can be closed.

PR #1135, opened 2022-12-06
PAYEE-984: Upgrade tai-frontend libraries to more recent and stable versions

vivekhmrc:
This is to avoid bobby violations

- [x] Unit tests passing
- [x] Coverage reached
- [ ] Reviewed


vivekhmrc:
There are incompatibilities among the library dependencies so the versions are selected to not break the dependencies

PR #1142, opened 2022-12-15
PAYEE-758: Remove TaiResponse from TaxAccountConnector .nonTaxCodeInc…

Jake-Raffe:
…omes

- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/DetailedIncomeTaxEstimateController.scala
+++ b/app/controllers/DetailedIncomeTaxEstimateController.scala
@@ -16,22 +16,17 @@
 
 package controllers
 
+import cats.implicits._
 import controllers.actions.ValidatePerson
 import controllers.auth.{AuthAction, AuthedUser}
-import javax.inject.{Inject, Singleton}
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
-import cats.implicits._
-
 import uk.gov.hmrc.renderer.TemplateRenderer
-import uk.gov.hmrc.tai.connectors.responses.TaiSuccessResponseWithPayload
 import uk.gov.hmrc.tai.model.TaxYear
-import uk.gov.hmrc.tai.model.domain.TaxAccountSummary
-import uk.gov.hmrc.tai.model.domain.income.{NonTaxCodeIncome, TaxCodeIncome}
-import uk.gov.hmrc.tai.model.domain.tax.TotalTax
 import uk.gov.hmrc.tai.service.{CodingComponentService, TaxAccountService}
 import uk.gov.hmrc.tai.viewModels.estimatedIncomeTax.DetailedIncomeTaxEstimateViewModel
 import views.html.estimatedIncomeTax.DetailedIncomeTaxEstimateView
 
+import javax.inject.{Inject, Singleton}
 import scala.concurrent.ExecutionContext
 import scala.util.control.NonFatal
 
@@ -61,7 +56,7 @@ class DetailedIncomeTaxEstimateController @Inject()(
           Right(taxCodeIncomes),
           taxAccountSummary,
           codingComponents,
-          TaiSuccessResponseWithPayload(nonTaxCodeIncome: NonTaxCodeIncome)
+          nonTaxCodeIncome
           ) =>
         implicit val user: AuthedUser = request.taiUser
         val model = DetailedIncomeTaxEstimateViewModel(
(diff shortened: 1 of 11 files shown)


app/uk/gov/hmrc/tai/service/TaxAccountSummaryService.scala
```diff
@@ -65,8 +64,8 @@ class TaxAccountSummaryService @Inject()(
           IncomeSources(livePensionIncomeSources, liveEmploymentIncomeSources, ceasedEmploymentIncomeSources),
           nonMatchingCeasedEmployments
         )
-      case (_, _, _, _, TaiUnauthorisedResponse(message), _) => throw new UnauthorizedException(message)
-      case _                                                 => throw new RuntimeException("Failed to fetch income details")
+    } recover {
+      case e: UnauthorizedException => throw e
```
brendanstraughan, line 68: not sure if we need this as we are throwing the exception in the connector

app/controllers/DetailedIncomeTaxEstimateController.scala
```diff
@@ -61,7 +56,7 @@ class DetailedIncomeTaxEstimateController @Inject()(
           Right(taxCodeIncomes),
           taxAccountSummary,
           codingComponents,
-          TaiSuccessResponseWithPayload(nonTaxCodeIncome: NonTaxCodeIncome)
+          nonTaxCodeIncome
```
brendanstraughan, line 59: We have a handler for any exceptions here, will this be enough?

app/controllers/EstimatedIncomeTaxController.scala
```diff
@@ -59,12 +59,12 @@ class EstimatedIncomeTaxController @Inject()(
...
       codingComponentService.taxFreeAmountComponents(nino, TaxYear()),
-      partialService.getIncomeTaxPartial
-    ).mapN {
+      partialService.getIncomeTaxPartial)
+      .mapN {
         case (
             taxAccountSummary,
             totalTaxDetails,
-            TaiSuccessResponseWithPayload(nonTaxCodeIncome: NonTaxCodeIncome),
+            nonTaxCodeIncome,
```
brendanstraughan, line 67: same here we have a recover here

brendanstraughan:
I think we can remove some code here, I think we can just handle the exceptions in one place, probably the controller? We are handling them all over here

app/uk/gov/hmrc/tai/connectors/TaxAccountConnector.scala
```diff
@@ -80,14 +78,13 @@ class TaxAccountConnector @Inject()(httpHandler: HttpHandler, servicesConfig: Se
...
 
-  def nonTaxCodeIncomes(nino: Nino, year: TaxYear)(implicit hc: HeaderCarrier): Future[TaiResponse] =
-    httpHandler.getFromApiV2(nonTaxCodeIncomeUrl(nino.nino, year)) map (
-      json => TaiSuccessResponseWithPayload((json \ "data").as[Incomes].nonTaxCodeIncomes)
-    ) recover {
-      case e: UnauthorizedException => TaiUnauthorisedResponse(e.getMessage)
+  def nonTaxCodeIncomes(nino: Nino, year: TaxYear)(implicit hc: HeaderCarrier): Future[NonTaxCodeIncome] =
+    httpHandler.getFromApiV2(nonTaxCodeIncomeUrl(nino.nino, year)).map { json =>
+      (json \ "data").as[Incomes].nonTaxCodeIncomes
+    } recover {
```
sergi-hmrc, line 84: There is a slight difference in behavior compared to the previous code. In this case, UnauthorizedExceptions will log a warning, whereas in the previous code this would not be done.

sergi-hmrc: the question is whether we want to warn that or not

app/controllers/DetailedIncomeTaxEstimateController.scala
sergi-hmrc: I don't think it's necessary as if it doesn't pattern match it will default to a MatchException which is a NonFatal one

sergi-hmrc: if `nonTaxCodeIncomes` fails it will be recovered in the `case NonFatal` so it should be fine

brendanstraughan (approved):
Approving, after discussing with @sergi-hmrc there are separate tickets to look at how we are handling exceptions

sergi-hmrc (approved):

PR #1150, opened 2022-12-16
PAYEE-985: Tax code change date added to banner

sergi-hmrc:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


PerrettJ4 (approved):
lgtm

app/views/WhatDoYouWantToDoTileView.scala.html
```diff
@@ -33,14 +35,9 @@
         ec: ExecutionContext)
 
 @optimizelyAudience = @{
-    if(model.showTaxCodeChangeTile()) {
-        Some("TaxCodeChangeUser")
-    } else {
-        None
-    }
+    model.maybeMostRecentTaxCodeChangeDate.map(_ => "TaxCodeChangeUser")
```
sergi-hmrc, line 38: I don't think this is used. Can someone correct me if wrong?

app/uk/gov/hmrc/tai/viewModels/WhatDoYouWantToDoViewModel.scala
```diff
@@ -16,49 +16,9 @@
...
-import scala.collection.immutable.ListMap
+import java.time.LocalDate
 
 case class WhatDoYouWantToDoViewModel(
   isCyPlusOneEnabled: Boolean,
-  hasTaxCodeChanged: Boolean = false,
   showJrsTile: Boolean,
-  taxCodeMismatch: Option[TaxCodeMismatch] = None) {
-
-  def showTaxCodeChangeTile(): Boolean =
```
brendanstraughan, line 31: just out of curiosity, was this moved to the controller as this is best practice?

test/controllers/WhatDoYouWantToDoControllerSpec.scala
```diff
@@ -227,7 +246,7 @@ class WhatDoYouWantToDoControllerSpec extends BaseSpec with JsoupMatchers with B
         status(result) mustBe OK
 
         doc.title() must include(Messages("your.paye.income.tax.overview"))
-        doc.body().toString must include(Messages("tai.WhatDoYouWantToDo.ChangedTaxCode"))
+        doc.body().toString must include(Messages("tai.WhatDoYouWantToDo.ChangedTaxCode", mostRecentTaxCodeChangeDate))
```
brendanstraughan, line 249: is there a check for this to no include taxCodeChange banner? If so is it worth adding one in?

test/uk/gov/hmrc/tai/viewModels/WhatDoYouWantToDoViewModelSpec.scala
```diff
@@ -1,158 +0,0 @@
...
-  def createViewModel(
-    isCyPlusOneEnabled: Boolean,
-    hasTaxCodeChanged: Boolean = false,
-    taxCodeMismatch: Option[TaxCodeMismatch] = None): WhatDoYouWantToDoViewModel =
-    WhatDoYouWantToDoViewModel(isCyPlusOneEnabled, hasTaxCodeChanged, false, taxCodeMismatch)
-
-  val mismatchedTaxCode = TaxCodeMismatchFactory.mismatchedTaxCodeComplex
-  val matchedTaxCode = TaxCodeMismatchFactory.matchedTaxCode
-
-  "showTaxCodeChangeTile" must {
```
brendanstraughan, line 38: can we add the showTaxCodeChangeTile tests back in as it looks like it is untested now, unless I have missed them somewhere?

app/views/WhatDoYouWantToDoTileView.scala.html
brendanstraughan: @masahmad im not aware of any need to have optimizely stuff in our code are you?

app/controllers/WhatDoYouWantToDoController.scala
```diff
@@ -92,29 +92,47 @@ class WhatDoYouWantToDoController @Inject()(
...
   }
 
+  private def showTaxCodeChangeTile(hasTaxCodeChanged: HasTaxCodeChanged): Boolean =
+    (hasTaxCodeChanged.changed, hasTaxCodeChanged.mismatch) match {
+      case (_, Some(mismatch)) if mismatch.confirmedTaxCodes.isEmpty => false
+      case (true, Some(TaxCodeMismatch(false, _, _)))                => true
+      case _                                                         => false
+    }
+
+  private def mostRecentTaxCodeChangeDate(nino: Nino, hasTaxCodeChanged: HasTaxCodeChanged)(
```
brendanstraughan, line 102: nitpick: the name of this isnt to clear for me, as this does the showTaxCodeChangeTile check as well I think maybe displayTileWithDate or something like this would make more sense. I think this would read better in the view particularly.

app/uk/gov/hmrc/tai/viewModels/WhatDoYouWantToDoViewModel.scala
sergi-hmrc: i used the case class a placeholder for data and nothing else. In such case to me the only important thing is to hold the maybeMostRecentTaxCodeChangeDate. As this is the model and must have the date now, I could have used the apply/factory method to generate the date but that also meant knowing an optional date that comes from a service call... I thought that would be overcomplicating things that should be kept as part of the controller (the controller is the one that interacts with service calls)

finally, that shouldn't have been a def as the attributes used are values that will never change...

test/uk/gov/hmrc/tai/viewModels/WhatDoYouWantToDoViewModelSpec.scala
sergi-hmrc: no problem

app/controllers/WhatDoYouWantToDoController.scala
sergi-hmrc: suggest a name and i'll change (although I think the name that should be changed is showTaxCodeChangeTile as this method just computes the mostRecentTaxCodeChangeDate)

brendanstraughan: I like showTileWithDate, feel free to ignore :)

sergi-hmrc: changed

test/uk/gov/hmrc/tai/viewModels/WhatDoYouWantToDoViewModelSpec.scala
sergi-hmrc: added

app/views/WhatDoYouWantToDoTileView.scala.html
sergi-hmrc: removed

brendanstraughan (approved):

PR #1160, opened 2023-01-03
PAYEE-746 - Removing exception

brendanstraughan:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/controllers/IncomeController.scala
+++ b/app/controllers/IncomeController.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 HM Revenue & Customs
+ * Copyright 2023 HM Revenue & Customs
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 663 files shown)


app/controllers/IncomeController.scala
```diff
@@ -214,7 +214,8 @@ class IncomeController @Inject()(
 
                     case _ => throw new RuntimeException(s"Not able to found employment with id $empId")
                   }
-                case _ => throw new RuntimeException("Exception while reading employment and tax code details")
+                case _ =>
+                  errorPagesHandler.internalServerError("Exception while reading employment and tax code details")
```
sergi-hmrc, line 218: should we log any of the errors that may have occurred?

sergi-hmrc (approved):

app/controllers/IncomeController.scala
PerrettJ4: I'm confused, so as a user where will I be directed?

brendanstraughan: they will just go to the standard internal server error page @PerrettJ4

brendanstraughan: @sergi-hmrc would we do this in a recover block or like how you have done yours?

sergi-hmrc: my bad this method logs... sorry

masahmad (approved):

sergi-hmrc (approved):

PR #1202, opened 2023-02-10
PAYEE-1133 - custom js , css to fix language selector 

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


PerrettJ4 (approved):

PerrettJ4:
So do other services still have to bug? 

And only our service displays the toggle correctly? If so we should notify, and also be made aware when their fix is implemented?

Jake-Raffe (approved):

PR #1206, opened 2023-02-14
PAYEE-1166 fix alignment issue

masahmad:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/views/TaxFreeAmountView.scala.html
+++ b/app/views/TaxFreeAmountView.scala.html
@@ -109,9 +109,7 @@
                         </p></li>
                     </ul>
 
-            <div class="print-this">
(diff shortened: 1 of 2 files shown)


app/views/TaxFreeAmountView.scala.html
```diff
@@ -149,4 +147,12 @@ <h2 class="govuk-heading-m heading-section hide-for-print">@Messages("tai.taxCod
 
     </div>
 
+    <div class="govuk-grid-row">
```
PerrettJ4, line 150: Why do you need this outer grid row when there is one child?

PerrettJ4: Resolved in call :)

PerrettJ4 (approved):

PR #1268, opened 2023-08-29
Ddcnl 7776

SandeepUmredkar:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/uk/gov/hmrc/tai/config/HmrcModule.scala
```diff
@@ -0,0 +1,29 @@
...
+import play.api.{Configuration, Environment}
+
+class HmrcModule extends Module {
+  override def bindings(environment: Environment, configuration: Configuration): Seq[Binding[_]] = {
+    val defaultBindings: Seq[Binding[_]] = Seq(
+      bind[ApplicationStartUp].toSelf.eagerly()
+    )
+    defaultBindings
+  }
+}
```
pascal22p, line 29: There is already TaiModule which can be used:
class TaiModule extends Module

app/views/WhatDoYouWantToDoTileView.scala.html
```diff
@@ -16,20 +16,26 @@
...
-        appConfig: ApplicationConfig
-)(implicit request: AuthenticatedRequest[_],
+        appConfig: ApplicationConfig)(implicit request: AuthenticatedRequest[_],
         messages: Messages,
-        user: controllers.auth.AuthedUser
-)
+        user: controllers.auth.AuthedUser)
+
+@cyPlusOneToggle = @{Await.result(featureFlagService.get(CyPlusOneToggle), Duration(appConfig.SCAWrapperFutureTimeout, SECONDS))}
+@incomeTaxHistoryToggle = @{Await.result(featureFlagService.get(IncomeTaxHistoryToggle), Duration(appConfig.SCAWrapperFutureTimeout, SECONDS))}
```
pascal22p, line 38: Don't do this. It is ok ish for sca wrapper because we are going to refactor it later.

Pass the value of the toggle to the view

test/views/html/WhatDoYouWantToDoTileViewSpec.scala
```diff
@@ -47,6 +54,11 @@ class WhatDoYouWantToDoTileViewSpec extends TaiViewSpec {
 
     "display cards correctly" when {
       "CY+1 is not enabled" in {
+        reset(mockFeatureFlagService)
```
pascal22p, line 57: Use beforeEach for this.
It should be where you bind the mock

pascal22p (changes requested):

app/views/WhatDoYouWantToDoTileView.scala.html
SandeepUmredkar: Done, Yeah I tried to keep all SCA thing similar, hence did it like this. but I agree that it should be passed to the view as like other flags.

pascal22p (dismissed):

pascal22p (approved):

PR #1324, opened 2024-07-08
DDCNL-8254

SandeepUmredkar:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/controllers/employments/UpdateEmploymentController.scala
```diff
@@ -160,86 +160,90 @@ class UpdateEmploymentController @Inject() (
...
+        form => {
+          val yesNoTask = form.yesNoChoice match {
             case Some(yn) if yn == FormValuesConstants.YesValue =>
-              mandatoryData ++ Map(UpdateEmploymentConstants.TelephoneNumberKey -> form.yesNoTextEntry.getOrElse(""))
-            case _ => mandatoryData ++ Map(UpdateEmploymentConstants.TelephoneNumberKey -> "")
+              journeyCacheNewRepository.set(
+                request.userAnswers
+                  .setOrException(
+                    UpdateEmploymentTelephoneQuestionPage,
+                    Messages(s"tai.label.${form.yesNoChoice.getOrElse(FormValuesConstants.NoValue).toLowerCase}")
```
grantrobson, line 177: Shouldn't this be YesValue?
Also lines 173-180 are almost identical to lines 182 - 190 so could make more concise by moving out of case match and parameterising

app/controllers/employments/UpdateEmploymentController.scala
```diff
@@ -160,86 +160,90 @@ class UpdateEmploymentController @Inject() (
...
+          Future.successful(
             Ok(
               updateEmploymentCheckYourAnswers(
                 UpdateEmploymentCheckYourAnswersViewModel(
-                  mandatoryJourneyValues.head.toInt,
-                  mandatoryJourneyValues(1),
-                  mandatoryJourneyValues(2),
-                  mandatoryJourneyValues(3),
-                  optionalSeq.head
+                  employmentId.get,
```
grantrobson, line 214: Don't use .gets here. Instead of using _ in case match at lines 209 say case (Some(empId), ...) and use these

app/controllers/employments/UpdateEmploymentController.scala
```diff
@@ -160,86 +160,90 @@ class UpdateEmploymentController @Inject() (
...
-                                                 )
-                                                 .getOrFail
-      model = IncorrectIncome(mandatoryCacheSeq(1), mandatoryCacheSeq(2), optionalCacheSeq.head)
-      _ <- employmentService.incorrectEmployment(user.nino, mandatoryCacheSeq.head.toInt, model)
-      _ <-
-        successfulJourneyCacheService
-          .cache(s"${TrackSuccessfulJourneyConstants.UpdateEndEmploymentKey}-${mandatoryCacheSeq.head}", true.toString)
-      _ <- journeyCacheService.flush()
-    } yield Redirect(controllers.employments.routes.UpdateEmploymentController.confirmation())
+    val employmentId = request.userAnswers.get(UpdateEmploymentIdPage)
```
grantrobson, line 230: Better to use case match here as above instead of .gets

grantrobson (approved):

pascal22p (approved):

PR #1326, opened 2024-08-01
DDCNL-7799 Moved AddPensionProviderController to new journey cache

Nj147:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


it/test/controllers/ContentsCheckSpec.scala
```diff
@@ -654,23 +655,33 @@ class ContentsCheckSpec extends IntegrationSpec with MockitoSugar with Matchers
...
-    .setOrException(AddEmploymentStartDatePage, LocalDate.of(2022, 7, 10))
-    .setOrException(UpdateEmploymentIdPage, 1)
-    .setOrException(UpdateEmploymentNamePage, "H M Revenue and Customs")
-    .setOrException(UpdateEmploymentTelephoneQuestionPage, "No")
-    .setOrException(UpdateEmploymentDetailsPage, "Details")
+  private val userAnswers = UserAnswers(
+    "",
+    "",
+    Json.obj(
+      EndEmploymentIdPage.toString                     -> 1,
```
grantrobson, line 662: These will work. But if the path on any of these objects changes in the future then they will fail. Using page.toString isn't really enough. If you look at setorexception you will see it stores the value in a concatenation of the path plus the toString.

grantrobson (dismissed):

grantrobson (approved):

PR #1366, opened 2025-05-13
Update routes

MrSleeth:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed

### DIFF
--- a/app/testOnly/connectors/TaiConnector.scala
+++ b/app/testOnly/connectors/TaiConnector.scala
@@ -33,20 +33,20 @@ class TaiConnector @Inject() (httpClientV2: HttpClientV2, servicesConfig: Servic
   def url(path: String): String = s"$serviceUrl$path"
 
   def employmentDetails(nino: String, taxYear: Int)(implicit hc: HeaderCarrier): Future[HttpResponse] = {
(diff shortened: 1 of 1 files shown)


conf/testOnlyDoNotUseInAppConf.routes
```diff
@@ -12,11 +12,11 @@
 # Add all the application routes to the prod.routes file
 # microservice specific routes
 
-GET        /check-income-tax/test-only/update-income/delete/:id                                                testOnly.controllers.TaiUpdateIncomeController.delete(id: Int)
+GET        /test-only/check-income-tax/test-only/update-income/delete/:id                                                testOnly.controllers.TaiUpdateIncomeController.delete(id: Int)
```
grantrobson, line 15: tai-fe calls the routes you changed in tai-be - so you need to change here too: https://github.com/hmrc/tai-frontend/blob/57b4ccceef14b556ac0ff71061708816e3470eb1/app/testOnly/connectors/TaiConnector.scala#L36 

I'm not sure whether you need to change the fe routes - are there warnings in tai-fe? If you do need to change them then could you remove the second test-only from the routes?

grantrobson (approved):

PR #1371, opened 2025-05-19
Ddcnl 10498

SandeepUmredkar:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/uk/gov/hmrc/tai/viewModels/TaxFreeAmountSummaryViewModel.scala
```diff
@@ -215,86 +215,100 @@ case class TaxFreeAmountSummaryRowViewModel(label: TaxSummaryLabel, value: Strin
 
 object TaxFreeAmountSummaryRowViewModel extends ViewModelHelper {
 
+  private val blockedAllowanceComponentTypes: Set[AllowanceComponentType] = Set(
```
grantrobson, line 218: Are we sure we got all of the items in google doc? For example I can't find TrustsSettlementsAndEstatesAtBasicRate in this PR

grantrobson: Mind, I can see that only the message for this one exists in the messages file. Can this be removed if it's not used?

SandeepUmredkar: So, the only values existing in the blocked set are the one which has been mentioned in the google doc and also there are components with that name and also the link would be taxFreeAllowanceLinkUrl. Apart from that, There are components which are not in the set
1) which are existing in excel but the link is different 
2) which are not existing in excel but the link is taxFreeAllowanceLinkUrl

grantrobson (approved):

PR #1393, opened 2025-08-29
Ddcnl 10536: refactor update estimate pay

SandeepUmredkar:
- [ ] Unit tests passing
- [ ] Coverage reached
- [ ] Reviewed


app/controllers/IncomeController.scala
```diff
@@ -333,38 +331,46 @@ class IncomeController @Inject() (
 
   def editPensionIncome(empId: Int): Action[AnyContent] = authenticate.authWithDataRetrieval.async { implicit request =>
     implicit val user: AuthedUser = request.taiUser
+    val nino                      = user.nino
+    val ua                        = request.userAnswers
+
+    val payToDateOpt = ua.get(UpdateIncomePayToDatePage)
```
grantrobson, line 337: What is this page? Where is it set? Normally each page object should correspond to an actual page, hence the name, ie a controller where the value is set, but this one doesn't seem to work that way. It's pretty confusing. We'd probably need to have a proper look at how the pages are used. Having said that, tai is going to be decommissioned and replaced by pega so maybe not worth the trouble

SandeepUmredkar: I think, this is left from the one where service call used to populate it. It does not seem to be coming from the page. So I will get rid of it and use the service call in place of this

pascal22p (changes requested):

pascal22p (changes requested):
Remove print statements

pascal22p (dismissed):

